PR #6, opened 2016-02-19
Modify the ArithOp that is selected when the arguments are floating-point values

robdockins:
So, e.g., we select `FAdd` instead of `Add` when the type of the arguments
are floating-point.


elliottt:
Thanks Rob!

PR #60, opened 2017-11-27
Fix parse of fast math field.

TomMD:
This "fix" doesn't appear to match my reading of the llvm C++ source (Bitcodereader.cpp) but our reading of the fast math flag appears incorrect. Without this fix we can not parse zstd bitcode.

@atomb and the saw team in general.  I need this patch if I am to analyze certain bitcodes.  While I feel the behavior is an improvement I am not confident this patch does "the right thing" in all cases.  Should we merge?  Should we find someone to take point w/ hours to back them? I'd rather not accept the current broken behavior (i.e. fail to merge and fail to fix).


atomb:
I'd say that if it's unambiguously less broken, we should merge it, and then open a ticket to account for making it completely correct.

PR #72, opened 2018-11-17
Parse global variables ahead of the rest of the module

siddharthist:
This is a necessary step in parsing the value table ahead of the rest of the module (because global variables [are pushed into the value table](https://github.com/siddharthist/llvm-pretty-bc-parser/blob/value-table-preparse/src/Data/LLVM/BitCode/IR/Globals.hs#L58)), which is a necessary step in parsing the metadata ahead of the rest of the module.

Not sure why `partialGlobals` didn't have a strictness annotation on it before, but it seems to work fine now.


siddharthist:
@atomb Let me know what you think about the additional strictness annotations here. I know that they're not always the way to the best performance, but don't have heuristics built up around when to use them.

PR #80, opened 2018-11-26
Better error messages for exceptions

siddharthist:
Exceptions now have strings describing what the error actually was.

This is helpful because e.g. there are several places where a `BadValueRef` can be thrown from, and it's hard to tell which occurred.


siddharthist:
Jenkins build after merging latest changes on `master`: https://build.galois.com/job/langston-llvm-disasm-quick-fuzz-branch/10/

PR #81, opened 2018-11-26
Remove outdated TODO about DILocation/MDLocation in LLVM 3.7+

siddharthist:
It appears this has been fixed in llvm-pretty:
- https://github.com/elliottt/llvm-pretty/blob/47aace66cdc5abc2294918b9665ef14b4d22762e/src/Text/LLVM/PP.hs#L46
- https://github.com/elliottt/llvm-pretty/commit/53773a3c34a87d3675994e1fe2c3cbd205c7d30a

Original commit: https://github.com/GaloisInc/llvm-pretty-bc-parser/commit/0f23af51c0df0e55afcce8a32ec8db49face966a

Does this seem right @TomMD @acfoltzer ?

### DIFF
--- a/src/Data/LLVM/BitCode/IR/Metadata.hs
+++ b/src/Data/LLVM/BitCode/IR/Metadata.hs
@@ -355,9 +355,6 @@ parseMetadataEntry vt mt pm (fromEntry -> Just r) =
 
   -- [distinct, line, col, scope, inlined-at?]
   7 -> label "METADATA_LOCATION" $ do
-    -- TODO: broken in 3.7+; needs to be a DILocation rather than an
-    -- MDLocation, but there appears to be no difference in the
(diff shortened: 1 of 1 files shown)


siddharthist:
From @TomMD:
> I looked.  I don't think of the change in the pretty printer as having any hold over the validity of the bc-parser package.  As I understand, and my memory is empty so this is just from my glancing, LLVM renamed MDLocation to DILocation from 3.6 to 3.7.  So we parse DILocation values (in llvm terms) but label the parse (for debugging purposes) as an MDLocation.  The comment can be deleted - there exists a synonym for historic reasons.  The pretty printer must take this into account since LLVM-as 3.7+ won't parse 'MDLocation' but what we call the tree element is immaterial.

PR #84, opened 2018-11-28
Add length assertions for various metadata records

siddharthist:
See commit messages for links to LLVM source.

This should help us fail quickly and with an informative message when encountering malformed bitcode (or when the implementation has gone awry!)

PR #87, opened 2018-11-30
Strict references to metadata strings

siddharthist:
I accidentally based this on #84, so it will have to be merged after that.

See the comment on the field `pmStrings`. Basically, we can be strict about these, so we should be. Additionally, this is necessary for solving an issue around `MDString`-based references in newer LLVM versions.


siddharthist:
Jenkins build: https://build.galois.com/job/langston-llvm-disasm-quick-fuzz-branch/12/

Issue #86, opened 2018-11-29
Roundtrip tests


siddharthist:
We currently have 3 very good test suites:
 - Fuzzing: Ensures that the parser can handle a decent subset of the LLVM emitted by `clang` on C programs
 - Regression: Ensures that the parser's behavior changes only when intended
 - disasm-tests: Ensure that the parser can handle some specific `.ll` files (with the option to compare the output with `llvm-dis`)

We could provide further semantic guarantees on our parser by adding a roundtrip test: 
 1. Disassemble some LLVM/Clang-generated `.bc` files with `llvm-disasm`, producing `.ll` files
 2. Re-assemble the `.ll` files using `llvm-as`, producing `.bc` files
 3. Disassemble the `.bc` files again with `llvm-disasm`

At the end of this process, we can use the following comparisons to provide additional guarantees:
 1. Is the re-assembled bitcode from step (2) identical to the original? If so, our parser/printer is bitcode semantics preserving (even though it does not output identical `.ll` files to `llvm-dis`)
 2. Are the ASTs resulting from steps (1) and (3) identical? If so, the printer correctly outputs all the data the parser tracks.
 3. Are the `.ll` files resulting from steps (1) and (3) identical? If so, our parser/printer is idempotent.

PR #89, opened 2018-12-01
Regression tests: allow cabal new-build

siddharthist:
Give users more flexibility on where they get their dependencies from. Also use `checkout` rather than `reset`, since users might include `llvm-pretty` as an optional package in their `cabal.project.local`.


siddharthist:
The latest commit also fixes #90 and #86

siddharthist:
Note that this commit disables the function `stripComments` which had buggy behavior (`{` and `}` were being removed from the output). This isn't a breaking change because the normalization is only for human eyes anyway.

PR #91, opened 2018-12-03
Roundtrip tests

### LINKED ISSUE #86: Roundtrip tests

siddharthist:
Fixes #86


siddharthist:
Diffing is happening as expected in CI: https://travis-ci.org/GaloisInc/llvm-pretty-bc-parser/jobs/463059443#L1583

PR #100, opened 2018-12-07
New-style (non-MDString-based) composite type references 

siddharthist:
This has been tested and found to fix parsing of a _very_ large C++ program. Fuzzing should tell us whether it's backwards compatible.

Jenkins build: https://build.galois.com/job/langston-llvm-disasm-quick-fuzz-branch/28/

The included clang++-generated bitcode only parses after this PR.


siddharthist:
Closing in favori of #101. It's actually not totally clear to me that we need this functionality.

Issue #121, opened 2019-02-18
Missing case for x86_fp80 constants


atomb:
The following C code leads to a parse error with the message `parseConstantEntry: Unsupported type X86_fp80`

~~~~ .c
extern void __VERIFIER_assume(int);
#include <math.h>
extern void __VERIFIER_error(void);
extern double __VERIFIER_nondet_double();
// all classification

int main()
{
  double d1 = __VERIFIER_nondet_double();
  double _d1 = __VERIFIER_nondet_double();
  d1=_d1;
  __VERIFIER_assume(isnormal(d1));
  if(!(!isnan(d1))) __VERIFIER_error();
  if(!(!isinf(d1))) __VERIFIER_error();
  if(!(isfinite(d1))) __VERIFIER_error();

  double d2 = __VERIFIER_nondet_double();
  double _d2 = __VERIFIER_nondet_double();
  d2=_d2;
  __VERIFIER_assume(isinf(d2));
  if(!(!isnormal(d2))) __VERIFIER_error();
  if(!(!isnan(d2))) __VERIFIER_error();


  double d3 = __VERIFIER_nondet_double();
  double _d3 = __VERIFIER_nondet_double();
  d3=_d3;
  __VERIFIER_assume(isnan(d3));
  if(!(!isnormal(d3))) __VERIFIER_error();
  if(!(!isinf(d3))) __VERIFIER_error();
  if(!(d3!=d3)) __VERIFIER_error();


  double d4 = __VERIFIER_nondet_double();
  double _d4 = __VERIFIER_nondet_double();
  d4=_d4;
  __VERIFIER_assume(isfinite(d4));
  if(!(!isnan(d4))) __VERIFIER_error();
  if(!(!isinf(d4))) __VERIFIER_error();


  double d5 = __VERIFIER_nondet_double();
  double _d5 = __VERIFIER_nondet_double();
  d5=_d5;
  __VERIFIER_assume(!isnan(d5) && !isinf(d5));
  if(!(isfinite(d5))) __VERIFIER_error();
}
~~~~

This is with the following `clang --version`:

~~~~
Apple LLVM version 10.0.0 (clang-1000.11.45.5)
Target: x86_64-apple-darwin18.2.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
~~~~

siddharthist:
[...]

PR #132, opened 2019-07-23
Fp80 support

kquick:
Resolves https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/121

Note that this does not use the long-double package as identified in that issue.  The primary reason for this was that the long-double package is very focused on FFI interaction, which has the advantage that C library fpFUNC() calls can be used to implement fp80 operations, but it also seems to come with some significant performance overheads/warnings.  The implementation here (and in llvm-pretty) just uses a simple datatype to store the exponent and significant values, leaving any semantic interpretation and manipulation to consumers.  If there is interest in switching to an implementation based on the  long-double package in the future it should be a very simple change to make.

### DIFF
--- a/src/Data/LLVM/BitCode/IR/Constants.hs
+++ b/src/Data/LLVM/BitCode/IR/Constants.hs
@@ -18,10 +18,10 @@ import           Control.Monad (mplus,mzero,foldM,(<=<))
 import           Control.Monad.ST (runST,ST)
 import           Data.Array.ST (newArray,readArray,MArray,STUArray)
 import           Data.Bits (shiftL,shiftR,testBit)
+import qualified Data.LLVM.BitCode.BitString as BitS
 import qualified Data.Map as Map
 import           Data.Maybe (fromMaybe, isJust)
-import           Data.Word (Word8,Word16, Word32,Word64)
-import qualified Data.LLVM.BitCode.BitString as BitS --  ( BitString(..), showBitString )
+import           Data.Word (Word16, Word32,Word64)
 
 #if __GLASGOW_HASKELL__ >= 704
 import           Data.Array.Unsafe (castSTUArray)
@@ -391,7 +391,7 @@ parseConstantEntry t (getTy,cs) (fromEntry -> Just r) =
       Integer 64         -> build ValInteger
       FloatType Float    -> build ValFloat
       FloatType Double   -> build ValDouble
-      FloatType X86_fp80 -> error "fp80buildData ty r cs getTy"
+      FloatType X86_fp80 -> error "TBD: similar to `fp80buildData ty r cs getTy`, but potentially applied iteratively"
       x                  -> Assert.unknownEntity "element type" x
 
   23 -> label "CST_CODE_INLINEASM" $ do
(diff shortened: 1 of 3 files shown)


src/Data/LLVM/BitCode/IR/Constants.hs
```diff
@@ -20,7 +20,8 @@ import           Data.Array.ST (newArray,readArray,MArray,STUArray)
 import           Data.Bits (shiftL,shiftR,testBit)
 import qualified Data.Map as Map
 import           Data.Maybe (fromMaybe, isJust)
-import           Data.Word (Word32,Word64)
+import           Data.Word (Word8,Word16, Word32,Word64)
+import qualified Data.LLVM.BitCode.BitString as BitS --  ( BitString(..), showBitString )
```
siddharthist:
```suggestion
import qualified Data.LLVM.BitCode.BitString as BitS
```

src/Data/LLVM/BitCode/IR/Constants.hs
```diff
@@ -383,13 +385,14 @@ parseConstantEntry t (getTy,cs) (fromEntry -> Just r) =
...
-      FloatType Float  -> build ValFloat
-      FloatType Double -> build ValDouble
-      x                -> Assert.unknownEntity "element type" x
+      Integer 8          -> build ValInteger
+      Integer 16         -> build ValInteger
+      Integer 32         -> build ValInteger
+      Integer 64         -> build ValInteger
+      FloatType Float    -> build ValFloat
+      FloatType Double   -> build ValDouble
+      FloatType X86_fp80 -> error "fp80buildData ty r cs getTy"
```
siddharthist: Can we make this error message a bit easier to read?

src/Data/LLVM/BitCode/IR/Constants.hs
```diff
@@ -477,3 +483,41 @@ cast x = do
...
+-- LGH/lib/Bitcode/Reader/BitcodeReader.cpp:2196-2202), using only 64
+-- bits of the first field and 16 bits of the second field, discarding
+-- the extra bits.  This is the result of this build operation; if
+-- this result is used semantically, it should be analyzed as per
+-- LGH/lib/Support/APFloat.cpp:3076-3108.
+
+fp80build :: Type -> Record -> [Typed PValue] -> Parse Type
+          -> Parse (Parse Type, [Typed PValue])
+fp80build ty r cs getTy =
+  do -- values <- parseField r 0 (fieldArray numeric)
```
siddharthist: Remove?

src/Data/LLVM/BitCode/IR/Types.hs
```diff
@@ -130,7 +130,7 @@ parseTypeBlockEntry (fromEntry -> Just r) = case recordCode r of
       rty:ptys -> addType (FunTy rty ptys va)
       _        -> fail "function expects a return type"
 
-  10 -> label "TYPE_CODE_X86_FP80" (addType (PrimType (FloatType Half)))
+  10 -> label "TYPE_CODE_FP_HALF" (addType (PrimType (FloatType Half)))
```
robdockins: This doesn't seem right...

robdockins: Oops, I misread this, nevermind

PR #136, opened 2020-01-17
GHC 8.8 updates

robdockins:
`MonadFail` updates, pretty much

### DIFF
--- a/src/Data/LLVM/BitCode/Parse.hs
+++ b/src/Data/LLVM/BitCode/Parse.hs
@@ -60,10 +60,8 @@ instance Monad Parse where
   {-# INLINE (>>=) #-}
   Parse m >>= f = Parse (m >>= unParse . f)
 
-  {-# INLINE fail #-}
-  fail = failWithContext
-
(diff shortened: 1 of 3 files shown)


src/Data/LLVM/BitCode/Parse.hs
```diff
@@ -145,13 +151,13 @@ setRelIds b = Parse $ do
   put $! ps { psValueTable = (psValueTable ps) { valueRelIds = b }}
 
 getRelIds :: Parse Bool
-getRelIds  = Parse $ do
```
langston-barrett: Why did all these change? Can we split that into a different commit?

robdockins: The `getLastLoc` change is actually related to MonadFail.  I needed the `fail` to occur in the `Parse` monad rather than the underlying monad, which doesn't have a `fail` instance.  I think it was always a bug to fail that way.

I did this change in `getRelIds` (and several other places) for uniformity, but maybe it would be better to leave it alone.

PR #137, opened 2020-08-17
Stricter parsing for numeric values

robdockins:
Check that the parsed value actually fits into the type of value we are parsing, and fail if it does not.

In the process of making this change, it became apparent that
some of the code for parsing floating-point values was incorrectly
interpreting the bitstrings.  This will be fixed in a subsequent
patch.

### DIFF
--- a/src/Data/LLVM/BitCode/IR/Constants.hs
+++ b/src/Data/LLVM/BitCode/IR/Constants.hs
@@ -392,9 +392,8 @@ parseConstantEntry t (getTy,cs) (fromEntry -> Just r) =
       Integer 16         -> build ValInteger
       Integer 32         -> build ValInteger
       Integer 64         -> build ValInteger
---      FloatType Float    -> build ValFloat
---      FloatType Double   -> build ValDouble
---      FloatType X86_fp80 -> error "TBD: similar to `fp80buildData ty r cs getTy`, but potentially applied iteratively"
+      FloatType Float    -> build (ValFloat . castFloat)
+      FloatType Double   -> build (ValDouble . castDouble)
       x                  -> Assert.unknownEntity "element type" x
 
   23 -> label "CST_CODE_INLINEASM" $ do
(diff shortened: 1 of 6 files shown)


src/Data/LLVM/BitCode/IR/Constants.hs
```diff
@@ -392,9 +392,9 @@ parseConstantEntry t (getTy,cs) (fromEntry -> Just r) =
       Integer 16         -> build ValInteger
       Integer 32         -> build ValInteger
       Integer 64         -> build ValInteger
-      FloatType Float    -> build ValFloat
-      FloatType Double   -> build ValDouble
-      FloatType X86_fp80 -> error "TBD: similar to `fp80buildData ty r cs getTy`, but potentially applied iteratively"
+--      FloatType Float    -> build ValFloat
```
robdockins, line 395: Note to self: still need to fix this.

robdockins:
I'm still waiting for the CI tests to come through and verify, but I think this is ready to go.

src/Data/LLVM/BitCode/IR/Metadata.hs
```diff
@@ -672,29 +739,29 @@ parseMetadataEntry vt mt pm (fromEntry -> Just r) =
...
-        <*> (mdForwardRefOrNull ctx mt <$> parseField r (adj 16) numeric) -- dispDeclaration
-        <*> (mdForwardRefOrNull ctx mt <$> parseField r (adj 17) numeric) -- dispVariables
-        -- Indices 18-19 seem unused.
-        <*> (optFwdRef hasThrownTypes 20)                                 -- dispThrownTypes
+            then parseField r (16 + offsetB) numeric
+            else return 0)                                                      -- dispThisAdjustment
+        <*> pure diFlags                                                        -- dispFlags
+        <*> pure isOptimized                                                    -- dispIsOptimized
+        <*> (optFwdRef hasUnit (12 + offsetB))                                  -- dispUnit
+        <*> (mdForwardRefOrNull ctx mt <$> parseField r (13 + offsetB) numeric) -- dispTemplateParams
```
kquick, line 762: In lines 755-759, the delta with `offsetB` against the old value is 3, but for lines 760-762, the delta is 2, then back to 3 again in line 763.  Just want to confirm that this is the intended set of offset base values.

robdockins: Yeah, this algorithm is nuts.  I double checked against the LLVM parser code, as far as I can tell, I've copied the logic accurately.

PR #141, opened 2020-10-23
Add support for some LLVM 11 & 12 metadata changes

atomb:
This depends on corresponding changes in `llvm-pretty` here: https://github.com/elliottt/llvm-pretty/pull/75

### DIFF
--- a/src/Data/LLVM/BitCode/IR/Metadata.hs
+++ b/src/Data/LLVM/BitCode/IR/Metadata.hs
@@ -573,7 +573,7 @@ parseMetadataEntry vt mt pm (fromEntry -> Just r) =
         (addDebugInfo isDistinct (DebugInfoDerivedType didt)) pm
 
     18 -> label "METADATA_COMPOSITE_TYPE" $ do
-      assertRecordSizeBetween 16 17
+      assertRecordSizeBetween 16 18
       ctx        <- getContext
       isDistinct <- parseField r 0 nonzero
       dict       <- DICompositeType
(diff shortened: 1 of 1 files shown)


src/Data/LLVM/BitCode/IR/Metadata.hs
```diff
@@ -627,20 +630,32 @@ parseMetadataEntry vt mt pm (fromEntry -> Just r) =
...
-      dicuDWOId <-
-        if recordSize <= 14
-        then pure 0
-        else parseField r 14 numeric
-      dicuSplitDebugInlining <-
-        if recordSize <= 16
-        then pure True
-        else parseField r 16 nonzero
-      let dicu' = dicu dicuDWOId dicuSplitDebugInlining
+        <*> (if recordSize <= 15
```
langston-barrett, line 633: Is there a reason that the 15 and 14 size cases don't appear in numerical order?

langston-barrett (approved):
I didn't track down the LLVM changes and ensure that this is the intended field naming/ordering, but I can if you'd like a second set of eyes there.

Thanks for making this happen!

src/Data/LLVM/BitCode/IR/Metadata.hs
atomb: Not a good one. It was just the minimal change: that's the order they were in before. But, since I'm changing all those lines anyway, I might as well put them in numerical order.

atomb: Oh, actually, there is a better reason: that's the order they were in before so that's the order they appear in the datatype in `llvm-pretty`. If we change that, we'd have to change all clients of the datatype.

langston-barrett: Gotcha. Thanks!

PR #142, opened 2020-10-23
Add workflow to fuzz with different LLVM versions

atomb:
This is very incomplete so far


atomb:
Merging now because we can't test this workflow until it exists on `master`.

PR #143, opened 2020-12-22
Add a GitHub workflow to fuzz against all LLVM versions we intend to support

atomb:
This updates the LLVM fuzzing workflow to check against LLVM versions from 3.5 to 11.

Making this work reliably required a few changes to the fuzzing harness.


fuzzing/Main.hs
```diff
@@ -41,7 +41,7 @@ import Text.XML.Light (Attr(..), Element, ppTopElement, unode, unqual)
 
 -- | The @clang@ executable and flags for a particular test
 -- configuration, e.g., @("clang-3.8", "-O -w -g")@
```
kquick, line 43: The comment should be updated  to match the type alias target.

.github/workflows/llvm-quick-fuzz.yml
```diff
@@ -11,40 +11,55 @@ jobs:
...
+      - shell: bash
+        run: ln -s $PWD/llvm/bin/clang $PWD/llvm/bin/clangy
+
+      - shell: bash
+        name: Run fuzzing
+        run: |
+          cabal exec -ffuzz fuzz-llvm-disasm -- -n 5 --junit-xml results.xml -o results --collapse -I /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include --clang-flags="-O -w" --clang-flags="-O -w -g" -c clangy
+          test $(grep "failures=\"0\"" results.xml | wc -l) -eq 2
+        env:
+          CSMITH_PATH: "/usr/local/Cellar/csmith/2.3.0/include/csmith-2.3.0"
```
kquick, line 68: Is there any way to automate this in case the `brew install csmith` gets a different version or the location changes?

Maybe something like `CSMITH_PATH: "ls $(type -p csmith)/../include/* | tr '\n' ':'` to check relative to the binary (untested, I don't know what the actual would be)?

kquick (approved):

fuzzing/Main.hs
atomb: Ah, good point. I'll do that.

.github/workflows/llvm-quick-fuzz.yml
atomb: Yeah, I was wondering about this, too. I don't like it being so fragile. I'll see what I can do to make it better.

jared-w (approved):

robdockins (commented):
I don't feel like I have enough context to review this, but I'll say it seems like a good idea.

Do we expect these checks to be running on the PR? It doesn't seem like they are.

jared-w:
> Do we expect these checks to be running on the PR? It doesn't seem like they are.

It's set on a timer to run nightly so it won't run until it's merged into the default branch.

PR #144, opened 2020-12-22
Don't crash on newer versions of METADATA_SUBRANGE

atomb:
We don't currently have any tools that use this metadata, so I haven't
bothered to correctly parse it yet. But now we don't immediately fail
upon encountering it. Also, llvm-disasm now allows LLVM versions up to
11.


src/Data/LLVM/BitCode/IR/Metadata.hs
```diff
@@ -508,8 +508,13 @@ parseMetadataEntry vt mt pm (fromEntry -> Just r) =
       fail "not yet implemented"
 
     13 -> label "METADATA_SUBRANGE" $ do
-      assertRecordSizeIn [3]
-      isDistinct     <- parseField r 0 nonzero
+      assertRecordSizeIn [3, 5]
+      field0     <- parseField r 0 unsigned
+      let isDistinct = field0 /= 0
+      --  format = field0 `shiftR` 1
+      -- TODO: use the format to determine how to parse remaining fields.
```
kquick, line 515: Lets make an issue to match/reference this TODO.

kquick (approved):

src/Data/LLVM/BitCode/IR/Metadata.hs
atomb: Working on one that I'll submit as soon as I've merged this PR, so that it can include a pointer to the appropriate code.

PR #146, opened 2020-12-22
Migrate from Travis CI to GitHub Actions

atomb:


.github/workflows/build.yml
```diff
@@ -0,0 +1,56 @@
...
+        os: [ubuntu-latest]
+        ghc: ["8.6.5", "8.8.4", "8.10.2"]
+    name: build - ${{ matrix.ghc }} - ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+
+      - name: Install LLVM (for running tests)
+        run: sudo apt-get -y install llvm
```
jared-w, line 23: This is not guaranteed to be a stable llvm version or the one that's first on the path. If it works, though, it works :)
But I would immediately be suspicious of it should things start failing due to llvm down the line (either from code changes or github runner changes)

.github/workflows/build.yml
```diff
@@ -0,0 +1,56 @@
...
+        with:
+          ghc-version: ${{ matrix.ghc }}
+
+      - uses: actions/cache@v2
+        name: Cache cabal store
+        with:
+          path: |
+            ${{ steps.setup-haskell.outputs.cabal-store }}
+            dist-newstyle
+          key: cabal-${{ runner.os }}-${{ matrix.ghc }}
```
jared-w, line 36: You really want `cabal-${{ runner.os}}-${{ matrix.ghc }}-${{ github.sha }}` here with the fallback key being `cabal-${{ runner.os}}-${{ matrix.ghc }}-` (note the ending `-`). Otherwise the cache will never update and build times will start to get longer as time goes on. Especially if dependencies are ever changed.

.github/workflows/build.yml
```diff
@@ -0,0 +1,56 @@
...
+
+      - shell: bash
+        name: Clone llvm-pretty
+        run: git clone https://github.com/elliottt/llvm-pretty
```
jared-w, line 40:
```yaml
- uses: actions/checkout@v2
  with:
    repository: elliottt/llvm-pretty
    path: llvm-pretty
```

will get you more performance (but it's not required by any means)

.github/workflows/build.yml
```diff
@@ -0,0 +1,56 @@
...
+
+      - shell: bash
```
jared-w, line 42: If you're only testing on linux, you can drop the `shell: bash` as it's the default. Even then, cabal has the same CLI on windows and linux so technically running it in powershell works fine.

jared-w (approved):

.github/workflows/build.yml
atomb: The fuzzing workflow I'm in the process of should help ensure that this part will work with any released version of LLVM, so I'm not really worried in this case. The use of LLVM in this particular workflow is minimal, and should be pretty robust to version shifts.

.github/workflows/build.yml
atomb: Oh, yikes. Will do!

.github/workflows/build.yml
atomb: I'll switch to that.

kquick: Why is the actions/checkout version faster?

PR #147, opened 2021-02-15
Retry curl on failure for llvm quick fuzz

jared-w:
We've been getting spurious failures from downloading LLVM. I suspect it might be rate limiting or something to that effect. So I'm going to try a naive retry and see if that works.


atomb (approved):
Seems worth a try!

PR #154, opened 2021-05-19
Check in llvm-pretty as a submodule

### LINKED ISSUE #153: CI: Vendor in llvm-pretty as a submodule
Currently, the CI workflows simply [clone the `llvm-pretty` repo from the most recent commit](https://github.com/GaloisInc/llvm-pretty-bc-parser/blob/8d4e4c661e33dfd9b87e45eb6f19d8ed823934cc/.github/workflows/build.yml#L39-L42). This is fragile, as this makes it possible for the `llvm-pretty` checkout to contain changes that `llvm-pretty-bc-parser` cannot yet handle (see #152 for a recent example of this). Instead, we should declare `llvm-pretty` as a submodule so that we can accurately track which commit to use.

RyanGlScott:
Fixes #153.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,11 +36,6 @@ jobs:
           key: cabal-${{ runner.os }}-${{ matrix.ghc }}-${{ github.sha }}
           restore-keys: cabal-${{ runner.os }}-${{ matrix.ghc }}-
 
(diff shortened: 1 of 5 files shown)


RyanGlScott:
Superseded by #159.

Issue #158, opened 2021-07-20
Parsing failures


kquick:
I have some candidate code (PX4_Autopilot) that is having parser decode failures with multiple symptoms:

1) Failure to decode an `fcmp` operation value.  The operation values defined are 0-15, but the value that llvm-pretty-bc-parser is attempting to resolve is 24.
2) ignoring the above by defaulting to `ftrue`, I'm seeing invalid operand values for the instruction.

Here's the llvm-dis output:
```
    %147 = load float, float* %146, align 4, !dbg !217760
    %148 = fmul nsz arcp float %147, 0x3F8457AE140000000, !dbg !217761
    %149 = fcmp nsz arcp ogt float %145, %148, !dbg !217769
```
for llvm-pretty-bc-parser, the llvm-disasm has problems on `%149`, where the default to `ftrue` still shows:
```
   %149 = fcmp true float %145, %147, !dbg ...`
```
Note the second operand is decoded as `%147` whereas llvm decodes to `%148`.

It's unknown if there are other parsing problems, this one just happened to fail the parser due to the lookup in (1) above and therefore be detectable.

kquick:
n.b., the llvm-disasm helpfully shows the `0x3F8457AE140000000` constant in register `%148` to be `1.0e-2`

RyanGlScott:
I think I'm encountering the same issue, or at least very similar one. To reproduce, compile the following file:

```c
double a = 0;
void b() {
  for (int c = 0; c < 2; c++)
    a++;
  if (a)
    a = 7;
}

int main() {
  b();
  return 0;
}
```

With:

```
$ clang -O1 test.c -S -emit-llvm -fno-discard-value-names
$ clang -O1 test.c -c -emit-llvm -fno-discard-value-names
```

This will produce the following bitcode with Clang 10:

```
; ModuleID = 'test.c'
source_filename = "test.c"
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
target triple = "x86_64-pc-linux-gnu"

@a = dso_local local_unnamed_addr global double 0.000000e+00, align 8

; Function Attrs: nofree norecurse nounwind uwtable
define dso_local void @b() local_unnamed_addr #0 {
entry:
  %0 = load double, double* @a, align 8, !tbaa !2
  br label %for.body

for.cond.cleanup:                                 ; preds = %for.body
[...]

PR #161, opened 2021-09-02
Clean up treatment of fast-math flags during parsing

RyanGlScott:
Currently, `llvm-pretty-bc-parser` skips over fast-math flags when parsing. However, it was doing so incorrectly when parsing `FUNC_CODE_INST_CMP2` (i.e., `fcmp` instructions). Per the LLVM source code [here](https://github.com/llvm/llvm-project/blob/b0391dfc737ede147e128fe877045f61fc5e4905/llvm/lib/Bitcode/Reader/BitcodeReader.cpp#L4342-L4377), the fast-math flags are expected to come after all of the other fields, but the code in `llvm-pretty-bc-parser` was assuming that they would appear in the middle. See #158.

I addresses this issue by parsing all of the other fields upfront and avoiding parsing the last part, which correspond to fast-math flags. While I was in town, I also made sure to add comments in every other place in the parser where we avoid parsing fast-math flags.


atomb:
I'm happy with this if it solves #158, but I think @kquick is the best person to check whether or not it does.

RyanGlScott:
I'd try to check the [PX4-Autopilot code](https://github.com/PX4/PX4-Autopilot) myself, but I can't even figure out how to build that repo with `clang`.

RyanGlScott:
In any case, this definitely fixes the bugs in https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/158#issuecomment-910958621 and https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/158#issuecomment-911678881, which are cropping up in SV-COMP benchmarks. In light of this, I'm going to go ahead and merge this. I'll leave #158 open as a reminder to check PX4_Autopilot.

PR #170, opened 2021-10-06
remove `Aligned` from `SubWord`

Ptival:
The `Aligned` constructor to `SubWord` is morally equivalent to `SubWord 0 0`.
While it is nice as a visual marker in code, it makes consuming `SubWord`s
often more verbose/redundant than it needs be.

This proposes removing it altogether, while preserving some readability with
the `aligned` value.  We could also add an `Aligned` pattern for `SubWord 0 _`,
but at the moment, we never actually need to pattern-match on `Aligned`.

It also makes it so `splitWord` and `getBitStringPartial` can take a `SubWord`
as input, rather than the length and word separately, which looked a bit out of
place.

### DIFF
--- a/src/Data/LLVM/BitCode/GetBits.hs
+++ b/src/Data/LLVM/BitCode/GetBits.hs
@@ -13,12 +13,11 @@ module Data.LLVM.BitCode.GetBits (
 
 import Data.LLVM.BitCode.BitString
 
-import Control.Applicative (Applicative(..),Alternative(..),(<$>))
+import Control.Applicative (Alternative(..))
 import Control.Arrow (first)
 import Control.Monad (MonadPlus(..),when,replicateM_)
 import Data.Bits (shiftR)
 import Data.ByteString (ByteString)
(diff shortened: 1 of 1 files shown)


src/Data/LLVM/BitCode/GetBits.hs
```diff
@@ -90,24 +89,25 @@ instance MonadPlus GetBits where
 -- files get padded to a 32-bit boundary.
 data SubWord
```
atomb, line 90: Now that there's only one constructor, it might be worth making `SubWord` a `newtype`, since it's probably quite central in the performance story.

atomb (approved):
This seems like a nice simplification. I wonder whether making `SubWord` a `newtype` could perhaps improve performance (but I'm not sure if it would).

atomb:
Ah, I was thinking of a `newtype` in record form with strict fields. But it may not actually be an improvement.

atomb:
Oh, you're right. I was misremembering the record syntax. Although you can use that syntax, it only allows one field.

Issue #149, opened 2021-03-20
Error when parsing Linux kernel compiled with LLVM 8


langston-barrett:
This was on commit c38b86b, with kernel 4.14.39 compiled with GLLVM+Clang 8, [like so](https://gist.github.com/langston-barrett/022a6f65f9d994a9661ecfc0645cff45).
```
invalid function symbol in BLOCKADDRESS record
from:
	CST_CODE_BLOCKADDRESS
	CONSTANTS_BLOCK
	@local_bh_enable
	FUNCTION_BLOCK
	FUNCTION_BLOCK_ID
	value symbol table
	MODULE_BLOCK
	Bitstream
```
[vmlinux.bc.zip](https://github.com/GaloisInc/llvm-pretty-bc-parser/files/6176523/vmlinux.bc.zip)

langston-barrett:
Same problem with LLVM/Clang 7.

PR #172, opened 2021-10-07
fix BLOCKADDRESS parsing

Ptival:
This is an attempt at fixing https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/149

This attempt is dependent upon the approval of https://github.com/elliottt/llvm-pretty/pull/87

I just started looking at the llvm-pretty-bc-parser code base, so I am mostly making educated guesses about everything and would appreciate some second pair of eyes.

The issue reported `invalid function symbol in BLOCKADDRESS record` seems to happen because of this line:

https://github.com/GaloisInc/llvm-pretty-bc-parser/blob/24dc69a4e718193d65874d84466a7daecf8d8356/src/Data/LLVM/BitCode/IR/Constants.hs#L389

which does not seem to match its LLVM counterpart at all, even going back to version 7.0:

https://github.com/llvm/llvm-project/blob/4856a9330ee01d30e9e11b6c2f991662b4c04b07/llvm/lib/Bitcode/Reader/BitcodeReader.cpp#L2565-L2566

A naive attempt at replacing the line with:
```haskell
    valref <- field 1 numeric
    val <- forwardRef ctx valref t
```
leads to an infinite loop caused by this line:
https://github.com/GaloisInc/llvm-pretty-bc-parser/blob/24dc69a4e718193d65874d84466a7daecf8d8356/src/Data/LLVM/BitCode/IR/Constants.hs#L391

My understanding is that this triggers the evaluation of `val` to extract its value, which triggers the resolution of the `forwardRef`, which at some point must throw some fixpoint computation in a loop.

So my guess was that this reference should not be resolved now. However, the `ConstBlockAddr` constructor expects a resolved `Symbol`, and I don't think we can get that without resolving the reference?

So I figured we might be able to just keep that expression as an expression, but we need some new constructor to hold that. I put it in the `Value` datatype back at https://github.com/elliottt/llvm-pretty/pull/87 

Doing so, I am able to parse that whole vmlinux.bc all the way to the end.

Checking the pretty-printing of my `ValueBlockAddr`, it looks like:
```
  call void (i64, i32) @__local_bh_enable_ip(i64 ptrtoint (i8* void()* @local_bh_enable.1986%1 to i64), i32 512)
                                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
which we may want to amend but I don't know what would be nice.

### DIFF
--- a/src/Data/LLVM/BitCode/IR/Constants.hs
+++ b/src/Data/LLVM/BitCode/IR/Constants.hs
@@ -14,14 +14,13 @@ import           Data.LLVM.BitCode.Record
 import           Text.LLVM.AST
 
 import qualified Codec.Binary.UTF8.String as UTF8 (decode)
-import           Control.Monad (mplus,mzero,foldM,(<=<))
+import           Control.Monad (mplus,mzero,foldM,(<=<), when)
 import           Control.Monad.ST (runST,ST)
 import           Data.Array.ST (newArray,readArray,MArray,STUArray)
 import           Data.Bits (shiftL,shiftR,testBit, Bits)
 import qualified Data.LLVM.BitCode.BitString as BitS
 import qualified Data.Map as Map
 import           Data.Maybe (fromMaybe, isJust)
-import           Data.Semigroup ( (<>) )
 import           Data.Word (Word16, Word32,Word64)
 
 #if __GLASGOW_HASKELL__ >= 704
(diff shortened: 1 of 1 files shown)


atomb (approved):
Looks reasonable to me, once the `llvm-pretty` changes are merged and CI passes.

PR #180, opened 2021-11-22
switch from cereal to binary

Ptival:
@glguy has expressed interest in retiring https://hackage.haskell.org/package/cereal in favor of https://hackage.haskell.org/package/binary 

This PR implements this change. It is mostly neutral, except for the type signature of
https://hackage.haskell.org/package/binary-0.8.9.0/docs/Data-Binary-Get.html#v:runGetOrFail
which returns leftover bytestring and offset that we now throw away.

This change might also be welcome in that we may want/need to use the `runGetIncremental` interface for parsing the file more lazily.

This may warrant a new version number.


travitch (approved):
Seems reasonable to me!

src/Data/LLVM/BitCode/Bitstream.hs
```diff
@@ -90,14 +90,21 @@ data Bitstream = Bitstream
   , bsEntries  :: [Entry]
   } deriving (Show)
 
+runGet :: BG.Get a -> S.ByteString -> Either String a
```
RyanGlScott, line 93: Could this version of `runGet` be more succinctly implemented in terms of `binary`'s [`runGetOrFail`](https://hackage.haskell.org/package/binary-0.10.0.0/docs/Data-Binary-Get.html#v:runGetOrFail)?

kquick (approved):
Looks good, and I like the possibility of more laziness/performance this may open up.

src/Data/LLVM/BitCode/Bitstream.hs
Ptival: `runGetOrFail` works over lazy bytestrings, this `runGet` works over strict bytestrings.

I can try and implement it by converting the bytestring, I'll see if this has a performance impact or not.

RyanGlScott: Thanks! And sorry to give you more work to do. Then again, I imagine that `llvm-pretty-bc-parser` isn't the only `cereal`-using library that uses `runGet`, so it's likely we'll have to tackle this elsewhere. If we do, having some performance data on this would be helpful.

src/Data/LLVM/BitCode/Bitstream.hs
```diff
@@ -91,13 +91,15 @@ data Bitstream = Bitstream
   } deriving (Show)
 
 parseBitstream :: S.ByteString -> Either String Bitstream
-parseBitstream  = C.runGet (runGetBits getBitstream)
+parseBitstream = bimap thd3 thd3
+                 . BG.runGetOrFail (runGetBits getBitstream)
+                 . L.fromChunks . (:[])
```
RyanGlScott, line 96: `bytestring`'s [`fromStrict`](https://hackage.haskell.org/package/bytestring-0.11.3.0/docs/Data-ByteString.html#v:fromStrict) function would be a more idiomatic way to convert from a strict to lazy `ByteString`. This function was introduced in `bytestring-0.10`, so you'd need to bump the lower version bounds in the `.cabal` file accordingly.

src/Data/LLVM/BitCode/Bitstream.hs
```diff
@@ -91,13 +91,15 @@ data Bitstream = Bitstream
...
 
 parseBitCodeBitstream :: S.ByteString -> Either String Bitstream
-parseBitCodeBitstream  = C.runGet (runGetBits getBitCodeBitstream)
+parseBitCodeBitstream = parseBitCodeBitstreamLazy . L.fromChunks . (:[])
```
RyanGlScott, line 99: Similarly, use `fromStrict` here.

src/Data/LLVM/BitCode/Bitstream.hs
```diff
@@ -19,15 +19,15 @@ module Data.LLVM.BitCode.Bitstream (
...
 import Control.Monad (unless,replicateM,guard)
 import Data.Bits (Bits, bitSizeMaybe, bit)
-import Data.Monoid (Monoid(..))
 import Data.Word (Word8,Word16,Word32)
+import qualified Data.Binary.Get as BG
 import qualified Data.ByteString as S
 import qualified Data.ByteString.Lazy as L
 import qualified Data.Map as Map
-import qualified Data.Serialize as C
+import Data.Bifunctor (bimap)
```
RyanGlScott, line 29: `Data.Bifunctor` was introduced in `base-4.8`, so make sure to bump the lower version bounds in the `.cabal` file accordingly.

RyanGlScott (approved):
LGTM, aside from some minor nitpicks.

kquick:
Checked performance via 5 runs of `llvm-disasm` on a 12MB bitcode file piping the output to `/dev/null`.  The runs take roughly 30s and the averaged `user` time delta between `cereal` and `binary` was a mere 0.056s (0.19%) improvement.

PR #182, opened 2022-03-24
Support parsing `FUNC_CODE_INST_ATOMICRMW` (59) function codes

### LINKED ISSUE #175: support `FUNC_CODE_INST_ATOMICRMW`

RyanGlScott:
What was previously named `FUNC_CODE_INST_ATOMICRMW` (code 38) is now named `FUNC_CODE_INST_ATOMICRMW_OLD`. The new `FUNC_CODE_INST_ATOMICRMW` (code 59), introduced in LLVM 13, is like `FUNC_CODE_INST_ATOMICRMW_OLD`, except that it has an explicit `valty` field to go along with its `val` field. To accommodate this, we augment `parseAtomicRMW` to check if we are dealing with the old or new function code, and if it is the latter, we parse a type-value pair instead of just a value. This requires rearranging the code a bit to get the indices correct, but nothing too drastic.

I've included a disassembly test and verified that it works both before and after LLVM 13 (i.e., with both forms of instruction code).

Fixes #175.

### DIFF
--- a/disasm-test/tests/atomicrmw.ll
+++ b/disasm-test/tests/atomicrmw.ll
@@ -1,0 +1,4 @@
+define void @atomicrmw(i32* %a, i32 %i) {
+    %b = atomicrmw add i32* %a, i32 %i acquire
+    ret void
+}
--- a/src/Data/LLVM/BitCode/IR/Function.hs
+++ b/src/Data/LLVM/BitCode/IR/Function.hs
@@ -767,8 +767,8 @@ parseFunctionBlockEntry _ t d (fromEntry -> Just r) = case recordCode r of
     notImplemented
 
   -- LLVM 6.0: [ptrty, ptr, val, operation, vol, ordering, ssid]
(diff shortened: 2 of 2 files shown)


kquick:
I'm a little worried that changing the name might break client code that still uses the old name.  Would it be better to give a new name to the newer implementation?

RyanGlScott:
By "changing the name", are you referring to the change from `label "FUNC_CODE_INST_ATOMICRMW"` to `label "FUNC_CODE_INST_ATOMICRMW_OLD"`? Note that the `label` function is just a thin wrapper around [`binary`'s `label`](https://hackage.haskell.org/package/binary-0.10.0.0/docs/Data-Binary-Get.html#v:label), which does nothing but add information to the error message that the user would see if the subsequent code failed to parse. That is to say, the behavior of the parser itself won't change. The important bit is the function code, and that remains unchanged.

kquick:
No worries now, that was the confirmation I was looking for.

langston-barrett (approved):

PR #195, opened 2022-09-08
Clarify and isolate the BitString interface

kquick:
This makes various modifications to BitString to clarify the usage, intent, and API, and to minimize that API.

Notably:
* `Int` was used for counts of both Bits and Bytes, so a newtype for each is introduced to clarify intent.
* Functions to access and manipulate the BitString are used universally to allow the internals to be decoupled from the interface.

This work is creating a basis for subsequent work on performance and memory footprint.

### DIFF
--- a/src/Data/LLVM/BitCode/BitString.hs
+++ b/src/Data/LLVM/BitCode/BitString.hs
@@ -6,7 +6,6 @@ module Data.LLVM.BitCode.BitString (
   , showBitString
   , fromBitString
   , bitStringValue
-  , maskBits
   , take, drop
   , NumBits, NumBytes, pattern Bits', pattern Bytes'
   , bitCount
--- a/src/Data/LLVM/BitCode/BitString.hs
+++ b/src/Data/LLVM/BitCode/BitString.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE PatternSynonyms #-}
+
 module Data.LLVM.BitCode.BitString (
     BitString(BitString)
   , toBitString
@@ -6,6 +8,10 @@ module Data.LLVM.BitCode.BitString (
   , bitStringValue
   , maskBits
   , take, drop
+  , NumBits, NumBytes, pattern Bits', pattern Bytes'
+  , bitCount
+  , addBitCounts
+  , subtractBitCounts
   ) where
 
 import Data.Bits ((.&.),(.|.),shiftL,shiftR,bit,bitSizeMaybe, Bits)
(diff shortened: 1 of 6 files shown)


src/Data/LLVM/BitCode/BitString.hs
```diff
@@ -13,26 +19,49 @@ import Numeric (showIntAtBase)
...
+{-# COMPLETE Bytes' #-}
+
+bitCount :: NumBits -> Int
+bitCount (NumBits n) = n
+
+addBitCounts :: NumBits -> NumBits -> NumBits
+addBitCounts (NumBits a) (NumBits b) = NumBits $ a + b
+
+subtractBitCounts :: NumBits -> NumBits -> NumBits
+subtractBitCounts (NumBits a) (NumBits b) = NumBits $ a - b
```
RyanGlScott, line 40: These look like the beginnings of a `Num` instance. Would it be easier just to define one and use that rather than these assorted top-level functions? (A similar question for `NumBytes` as well.)

src/Data/LLVM/BitCode/GetBits.hs
```diff
@@ -88,33 +88,35 @@ instance MonadPlus GetBits where
...
 -- | @getBitString n@ grabs a @BitString@ of length @n@ from the next incoming
 -- word, yielding the remainder partial word.  On @n@ = @0@, it does not
 -- actually consume the next word.  Should not be called to read more than one
 -- 32-bit word at a time (will fail on @n@ > 32).
-getBitString :: Int -> BG.Get (BitString, SubWord)
-getBitString 0 = return (mempty, aligned)
-getBitString n | n > 32 =
+getBitString :: NumBits -> BG.Get (BitString, SubWord)
+getBitString (Bits' 0) = return (mempty, aligned)
+getBitString n | n > (Bits' 32) =
```
RyanGlScott, line 110:
```suggestion
getBitString n | n > Bits' 32 =
```

src/Data/LLVM/BitCode/GetBits.hs
```diff
@@ -88,33 +88,35 @@ instance MonadPlus GetBits where
...
+getBitString n = getBitStringPartial n . SubWord (Bits' 32) =<< BG.getWord32le
 
 -- | @getBitStringPartial n sw@ returns a @BitString@ of length @n@ from either
 -- the current subword @sw@ (with some @l@ bits available), potentially also
 -- reading the next incoming word if @n@ > @l@.  Should not be called to read
 -- more than one 32-bit word at a time (will fail on @n@ > 32).
-getBitStringPartial :: Int -> SubWord -> BG.Get (BitString, SubWord)
-getBitStringPartial n _ | n > 32 =
+getBitStringPartial :: NumBits -> SubWord -> BG.Get (BitString, SubWord)
+getBitStringPartial n _ | n > (Bits' 32) =
```
RyanGlScott, line 119:
```suggestion
getBitStringPartial n _ | n > Bits' 32 =
```

src/Data/LLVM/BitCode/BitString.hs
kquick: These are actually going away in an upcoming PR.  It's a good suggestion if I were keeping them, but the process here was  to shift from standard classes to explicit functions to track the usage preparatory to more substantial changes.

RyanGlScott: Gotcha.

RyanGlScott (approved):

PR #196, opened 2022-09-09
Explicit BitString interface and removal of internal SubWord

kquick:
This is another step in the work on performance improvements in this parser.  

* This PR makes the `BitString` API more explicit (removing some class interfaces such as `Monoid` and `Semigroup` in favor of explicit functions to clarify the API footprint).  
* Additionally, this refactors the internal implementation for `BitString`.  The previous implementation used `Data.Binary` exclusively to extract a 32-bit integer value from the `ByteString`.  This 32-bit integer was placed into a `SubWord` structure, from which individual bits were extracted until the `SubWord` became "empty", in which case it would be refilled with another 32-bit integer value obtained from `Data.Binary`.  This PR changes the `BitString` to directly keep track of the offset in the underlying `ByteString` (this offset was previously internal to the `Data.Binary` monad), and any value retrievals are performed directly from the `ByteString`, removing the `SubWord` layer of abstraction and manipulation.

A follow-on PR will take additional steps to optimize this implementation.

### DIFF
--- a/src/Data/LLVM/BitCode/BitString.hs
+++ b/src/Data/LLVM/BitCode/BitString.hs
@@ -1,23 +1,28 @@
 {-# LANGUAGE PatternSynonyms #-}
 
-module Data.LLVM.BitCode.BitString (
+module Data.LLVM.BitCode.BitString
+  (
     BitString(BitString)
+  , emptyBitString
   , toBitString
   , showBitString
   , fromBitString
   , bitStringValue
   , take, drop
+  , joinBitString
   , NumBits, NumBytes, pattern Bits', pattern Bytes'
   , bitCount
   , addBitCounts
   , subtractBitCounts
-  ) where
+  )
+where
 
 import Data.Bits ((.&.),(.|.),shiftL,shiftR,bit,bitSizeMaybe, Bits)
-import Data.Semigroup
 import Numeric (showIntAtBase)
 
 import Prelude hiding (take,drop,splitAt)
+
+----------------------------------------------------------------------
 
 newtype NumBits = NumBits Int deriving (Show, Eq, Ord)
 newtype NumBytes = NumBytes Int deriving (Show, Eq, Ord)
@@ -42,25 +47,40 @@ subtractBitCounts (NumBits a) (NumBits b) = NumBits $ a - b
 data BitString = BitString
   { bsLength :: !NumBits
   , bsData   :: !Integer
-  } deriving Show
+  } deriving (Show, Eq)
 
-instance Eq BitString where
-  BitString n i == BitString m j = n == m && i == j
+-- | Create an empty BitString
 
-instance Semigroup BitString where
-  BitString n@(NumBits n') i <> BitString m j =
-    BitString (addBitCounts n m) (i .|. (j `shiftL` n'))
+emptyBitString :: BitString
+emptyBitString = BitString (NumBits 0) 0
 
-instance Monoid BitString where
-  mempty = BitString (NumBits 0) 0
-  mappend = (<>)
+
+-- | Join two BitString representations together to form a single larger
+-- BitString.  The first BitString is the "lower" value portion of the resulting
+-- BitString.
+
+joinBitString :: BitString -> BitString -> BitString
+joinBitString a b =
+  let bitSizeA = bitCount $ bsLength a
+  in BitString { bsLength = addBitCounts (bsLength a) (bsLength b)
+               , bsData = bsData a .|. (bsData b `shiftL` bitSizeA)
+               }
+
 
 -- | Given a number of bits to take, and an @Integer@, create a @BitString@.
+
 toBitString :: NumBits -> Integer -> BitString
 toBitString len val = BitString len (val .&. maskBits len)
 
+
+-- | Extract the referenced Integer value from a BitString
+
 bitStringValue :: BitString -> Integer
 bitStringValue = bsData
+
+
+-- | Extract a target (Num) value of the desired type from a BitString (using
+-- fromInteger to perform the target type conversion).
 
 fromBitString :: (Num a, Bits a) => BitString -> a
 fromBitString (BitString l i) =
@@ -76,6 +96,7 @@ fromBitString (BitString l i) =
  where
  x    = fromInteger ival
  ival = i .&. maskBits l
+
 
 showBitString :: BitString -> ShowS
 showBitString bs = showString padding . showString bin
@@ -93,14 +114,22 @@ maskBits (NumBits len)
   | len <= 0  = 0
   | otherwise = pred (bit len)
 
+
+-- | Extract a smaller BitString with the specified number of bits from the
+-- "start" of a larger BitString.
+
 take :: NumBits -> BitString -> BitString
 take n bs@(BitString l i)
   | n >= l    = bs
   | otherwise = toBitString n i
 
+
+-- | Remove the specified number of bits from the beginning of a BitString and
+-- return the remaining as a smaller BitString.
+
 drop :: NumBits -> BitString -> BitString
 drop n (BitString l i)
-  | n >= l    = mempty
+  | n >= l    = emptyBitString
   | otherwise = BitString
                 (NumBits $ bitCount l - bitCount n)
                 (i `shiftR` (bitCount n))
--- a/src/Data/LLVM/BitCode/BitString.hs
+++ b/src/Data/LLVM/BitCode/BitString.hs
@@ -11,7 +11,7 @@ module Data.LLVM.BitCode.BitString
   , take, drop
   , joinBitString
   , NumBits, NumBytes, pattern Bits', pattern Bytes'
-  , bitCount
+  , bitCount, bitsToBytes, bytesToBits
   , addBitCounts
   , subtractBitCounts
   )
@@ -40,15 +40,19 @@ pattern Bytes' n = NumBytes n
 bitCount :: NumBits -> Int
 bitCount (NumBits n) = n
 
+{-# INLINE addBitCounts #-}
 addBitCounts :: NumBits -> NumBits -> NumBits
 addBitCounts (NumBits a) (NumBits b) = NumBits $ a + b
 
+{-# INLINE subtractBitCounts #-}
 subtractBitCounts :: NumBits -> NumBits -> NumBits
 subtractBitCounts (NumBits a) (NumBits b) = NumBits $ a - b
 
+{-# INLINE bytesToBits #-}
 bitsToBytes :: NumBits -> (NumBytes, NumBits)
 bitsToBytes (NumBits n) = (NumBytes $ n `shiftR` 3, NumBits $ n .&. 7)
 
+{-# INLINE bitsToBytes #-}
 bytesToBits :: NumBytes -> NumBits
 bytesToBits (NumBytes n) = NumBits $ n `shiftL` 3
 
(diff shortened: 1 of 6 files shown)


src/Data/LLVM/BitCode/BitString.hs
```diff
@@ -33,35 +40,62 @@ pattern Bytes' n = NumBytes n
...
 addBitCounts :: NumBits -> NumBits -> NumBits
 addBitCounts (NumBits a) (NumBits b) = NumBits $ a + b
 
+{-# INLINE subtractBitCounts #-}
 subtractBitCounts :: NumBits -> NumBits -> NumBits
 subtractBitCounts (NumBits a) (NumBits b) = NumBits $ a - b
 
+{-# INLINE bytesToBits #-}
+bitsToBytes :: NumBits -> (NumBytes, NumBits)
+bitsToBytes (NumBits n) = (NumBytes $ n `shiftR` 3, NumBits $ n .&. 7)
```
RyanGlScott, line 53: Presumably, the use of `shiftR` here will give strange results if `n` is negative? I'm guessing that there is an invariant that `NumBits` is non-negative. If so, it would be worth commenting that near `NumBits`' definition. (Similarly for `NumBytes`.)

src/Data/LLVM/BitCode/BitString.hs
```diff
@@ -33,35 +40,62 @@ pattern Bytes' n = NumBytes n
...
-  BitString n@(NumBits n') i <> BitString m j =
-    BitString (addBitCounts n m) (i .|. (j `shiftL` n'))
 
-instance Monoid BitString where
-  mempty = BitString (NumBits 0) 0
-  mappend = (<>)
+-- | Join two BitString representations together to form a single larger
+-- BitString.  The first BitString is the "lower" value portion of the resulting
+-- BitString.
+
```
RyanGlScott, line 75: A minor thing, but why put whitespace between a function and its Haddocks? It's somewhat odd to see them separated like this, especially when the opposite convention is so popular.

src/Data/LLVM/BitCode/BitString.hs
```diff
@@ -33,35 +40,62 @@ pattern Bytes' n = NumBytes n
...
 
-instance Semigroup BitString where
-  BitString n@(NumBits n') i <> BitString m j =
-    BitString (addBitCounts n m) (i .|. (j `shiftL` n'))
 
-instance Monoid BitString where
-  mempty = BitString (NumBits 0) 0
-  mappend = (<>)
+-- | Join two BitString representations together to form a single larger
+-- BitString.  The first BitString is the "lower" value portion of the resulting
```
RyanGlScott, line 73:
```suggestion
-- BitString.  The first BitString is the \"lower\" value portion of the resulting
```

src/Data/LLVM/BitCode/BitString.hs
```diff
@@ -93,14 +128,22 @@ maskBits (NumBits len)
   | len <= 0  = 0
   | otherwise = pred (bit len)
 
+
+-- | Extract a smaller BitString with the specified number of bits from the
+-- "start" of a larger BitString.
```
RyanGlScott, line 133:
```suggestion
-- \"start\" of a larger BitString.
```

src/Data/LLVM/BitCode/GetBits.hs
```diff
@@ -11,68 +11,106 @@ module Data.LLVM.BitCode.GetBits (
...
+import           Control.Monad.Fail ( MonadFail )
 import qualified Control.Monad.Fail
 #endif
 
 -- Bit-level Parsing -----------------------------------------------------------
 
-newtype GetBits a = GetBits { unGetBits :: SubWord -> BG.Get (a,SubWord) }
+newtype GetBits a = GetBits { unGetBits :: Either String (BitPosition -> BS.ByteString -> (BitsGetter a, BitPosition)) } -- Left is outer fail
+
+newtype BitPosition = BitPosition (NumBits, NumBits) -- (current, limit)
```
RyanGlScott, line 31: I think I would prefer `BitPosition` to be a data type with two fields rather than a newtype around a pair, as the latter is slightly more indirect.

src/Data/LLVM/BitCode/GetBits.hs
```diff
@@ -11,68 +11,106 @@ module Data.LLVM.BitCode.GetBits (
...
 -- | Run a @GetBits@ action, returning its value, and the number of bits offset
 -- into the next byte of the stream.
-runGetBits :: GetBits a -> BG.Get a
-runGetBits m = fst `fmap` unGetBits m aligned
+runGetBits :: GetBits a -> ByteString -> Either String a
+runGetBits m bs =
+  let startPos = BitPosition ( Bits' 0
+                             , bytesToBits $ Bytes' $ BS.length bs
+                             )
+  in ($ bs) =<< (fst . ($ bs) . ($ startPos) <$> unGetBits m)
```
RyanGlScott, line 43: I have a really hard time reading this code due to the excessive use of pointfree code.

src/Data/LLVM/BitCode/GetBits.hs
```diff
@@ -11,68 +11,106 @@ module Data.LLVM.BitCode.GetBits (
...
-    (y,off2) <- unGetBits x off1
-    return (g y,off2)
+  f <*> x =
+    GetBits $ do k <- unGetBits f
+                 l <- unGetBits x
+                 return $ \pos inp -> let (g,n) = k pos inp
+                                          (y,m) = l n inp
+                                          in ( \_ -> do g' <- g inp
+                                                        y' <- y inp
+                                                        return $ g' y'
```
RyanGlScott, line 65: Perhaps I'm not getting what the argument to `BitsGetter` represents, but I'm surprised to see this `BitsGetter` completely throw away its `ByteString` argument. What is the intuition behind this?

src/Data/LLVM/BitCode/GetBits.hs
```diff
@@ -82,96 +120,120 @@ instance MonadPlus GetBits where
...
-getByteString (Bytes' n) = do
-  bs <- BG.getByteString n
-  replicateM_ ((- n) `mod` 4) skipZeroByte
-  return (bs, aligned)
+-- | Extracts an Integer value of the specified number of bits from a ByteString,
+-- starting at the indicated bit position (fails with Left if the range to
+-- extract is not valid... i.e. > bitLimit).  Returns the Integer value along
+-- with the bit position following the extraction.
+
+extractFromByteString :: NumBits -> NumBits -> NumBits -> ByteString
```
RyanGlScott, line 128: Worth including some Haddock comments here to distinguish all of the `NumBits` arguments.

src/Data/LLVM/BitCode/GetBits.hs
```diff
@@ -82,96 +120,120 @@ instance MonadPlus GetBits where
...
+                      -> Either String (Integer, NumBits)
+extractFromByteString bitLimit startBit numBits bs =
+  let Bytes' s8 = fst (bitsToBytes startBit)
+      Bytes' r8 = fst (bitsToBytes numBits)
+      rcnt = r8 + 2 -- 2 == pre-shift overflow byte on either side
+      ws = BS.take rcnt $ BS.drop s8 bs
+      wi = BS.foldr (\w a -> a `shiftL` 8 .|. fromIntegral w) (0::Integer) ws
+      mask = ((1::Integer) `shiftL` bitCount numBits) - 1
+      vi = wi `shiftR` (bitCount startBit .&. 7) .&. mask
```
RyanGlScott, line 137: There's a lot of tricky bit-shifting code going on here—worth adding some comments to explain what is going on at a slightly higher level.

src/Data/LLVM/BitCode/BitString.hs
```diff
@@ -33,35 +40,62 @@ pattern Bytes' n = NumBytes n
...
+emptyBitString :: BitString
+emptyBitString = BitString (NumBits 0) 0
 
-instance Semigroup BitString where
-  BitString n@(NumBits n') i <> BitString m j =
-    BitString (addBitCounts n m) (i .|. (j `shiftL` n'))
 
-instance Monoid BitString where
-  mempty = BitString (NumBits 0) 0
-  mappend = (<>)
```
RyanGlScott, line 56: I'm not sure how to feel about removing these instances, as they seem quite useful. Do later changes prevent these from being feasible?

RyanGlScott (commented):
I haven't understood every line of code involved, but I like the direction of travel.

src/Data/LLVM/BitCode/BitString.hs
kquick: It's a readability habit I have since I find it helps me see the function declaration and corresponding types more easily because it isn't tucked right up against the docs.  That's just the background: I'm not strongly committed to this though, so I'll try to shift back towards the convention going forward.

src/Data/LLVM/BitCode/BitString.hs
kquick: It would cause issues, and although this isn't any different that the previous code it's reasonable to enforce some invariants around this (or even change the core types).  I'll add these in a future PR (not yet written) if that's OK.

src/Data/LLVM/BitCode/GetBits.hs
kquick: The base purpose of effort I'm undertaking here is to improve performance and memory utilization, so while in general I am in agreement with avoiding the use of "bare" tuples,  I've been trying to eliminate constructors (yes, this will currently have a tuple constructor, aka `,`, but in a future PR this will change to an unlifted type alias (`type BitPosition = (# ... #)`).  Are you OK with deferring your preference for the moment given this context?

src/Data/LLVM/BitCode/GetBits.hs
kquick: Noted.  This particular line will change in an upcoming PR, but I'll also keep the general concern in mind going forward.

src/Data/LLVM/BitCode/GetBits.hs
RyanGlScott: Ah, I didn't realize that this was fated to become an unboxed tuple. In that case, I'm fine with this approach.

src/Data/LLVM/BitCode/GetBits.hs
kquick: What you are seeing here is actually an awkward window into the evolution of this code.  The initial approach was attempting to defer the need for the ByteString to the inner portion (i.e. I wanted `newtype GetBits a = GetBits { unGetBits :: Either String (BitPosition -> (BitsGetter a, BitPosition) ) }` instead of the implementation in this PR) but in order to implement `Alternative` `<|>`, it was necessary to have the `ByteString` as an argument to `unGetBits` itself.  Thus, the `\_` on line 63 is the same value as `inp` (introduced at the outer layer on line 61).  This PR does not yet contain the refactoring to remove these duplicate inner arguments (that is coming in one of the future PR's).

As noted in a side-band communication but worth explicitly documenting here in this context: this is part of a series of PR's that iterate through a set of work designed to improve performance and reduce memory consumption.  They are being introduced as individual PR's to capture the different phases of work and provide PR-level higher level information about the general intent of those phases.  This leads to the unfortunately frequent use of "this will be changed in a future PR"; I apologize for the frequency of these uses and the corresponding request for deferment of judgement, but I think this process is *valuable* because it's easier to review smaller steps than a single massive PR, and I am committed to returning to these individual comments made during the process once we reach the endpoint (wherever that is) and re-address them.

src/Data/LLVM/BitCode/BitString.hs
kquick: Yes, later changes make these irrelevant, and removing these instances helped to explicitly clarify the API boundary.  If these reasons weren't in play, I would agree that keeping these would be useful.

RyanGlScott (approved):

PR #197, opened 2022-09-09
Improve performance and save space by using unlifted types

kquick:
Boxed types are normally convenient, but for high speed processing of items of known sizes, unboxed types consume less memory and can be much quicker to work with.  This PR also removes the duplication in using the input `ByteString` argument at both the outer and inner "layers" of the `GetBits` monad (that was noticed in the previous PR).

Note: a follow-on PR will shift the use of `Int#` in this PR to `Word#` to resolve the signed/unsigned issue.

### DIFF
--- a/llvm-pretty-bc-parser.cabal
+++ b/llvm-pretty-bc-parser.cabal
@@ -75,7 +75,10 @@ Library
 Executable llvm-disasm
   Main-is:             LLVMDis.hs
   Default-language:    Haskell2010
-  Ghc-options:         -Wall -Wunbanged-strict-patterns -O2
+  Ghc-options:         -Wall
+                       -Wunbanged-strict-patterns
+                       -O2
+                       -funbox-strict-fields
   Hs-source-dirs:      llvm-disasm
   Build-depends:       array                 >= 0.3,
                        base,
--- a/llvm-pretty-bc-parser.cabal
+++ b/llvm-pretty-bc-parser.cabal
@@ -56,6 +56,9 @@ Library
                        Data.LLVM.BitCode.Record
 
   Ghc-options:         -Wall
+                       -Wunbanged-strict-patterns
+                       -O2
+                       -funbox-strict-fields
 
   Build-depends:       array       >= 0.3,
                        base        >= 4.8 && < 5,
@@ -72,7 +75,7 @@ Library
 Executable llvm-disasm
   Main-is:             LLVMDis.hs
   Default-language:    Haskell2010
-  Ghc-options:         -Wall
+  Ghc-options:         -Wall -Wunbanged-strict-patterns -O2
   Hs-source-dirs:      llvm-disasm
   Build-depends:       array                 >= 0.3,
                        base,
--- a/src/Data/LLVM/BitCode/BitString.hs
+++ b/src/Data/LLVM/BitCode/BitString.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE MagicHash #-}
 {-# LANGUAGE PatternSynonyms #-}
 
 module Data.LLVM.BitCode.BitString
@@ -11,13 +13,15 @@ module Data.LLVM.BitCode.BitString
   , take, drop
   , joinBitString
   , NumBits, NumBytes, pattern Bits', pattern Bytes'
-  , bitCount, bitsToBytes, bytesToBits
+  , bitCount, bitCount#
+  , bitsToBytes, bytesToBits
   , addBitCounts
   , subtractBitCounts
   )
 where
 
-import Data.Bits ((.&.),(.|.),shiftL,shiftR,bit,bitSizeMaybe, Bits)
+import Data.Bits ( bit, bitSizeMaybe, Bits )
+import GHC.Exts
 import Numeric ( showIntAtBase, showHex )
 
 import Prelude hiding (take,drop,splitAt)
@@ -40,21 +44,26 @@ pattern Bytes' n = NumBytes n
 bitCount :: NumBits -> Int
 bitCount (NumBits n) = n
 
+bitCount# :: NumBits -> Int#
+bitCount# (NumBits (I# n#)) = n#
+
 {-# INLINE addBitCounts #-}
 addBitCounts :: NumBits -> NumBits -> NumBits
-addBitCounts (NumBits a) (NumBits b) = NumBits $ a + b
+addBitCounts (NumBits (I# a#)) (NumBits (I# b#)) = NumBits (I# (a# +# b#))
 
 {-# INLINE subtractBitCounts #-}
 subtractBitCounts :: NumBits -> NumBits -> NumBits
-subtractBitCounts (NumBits a) (NumBits b) = NumBits $ a - b
+subtractBitCounts (NumBits (I# a#)) (NumBits (I# b#)) = NumBits (I# (a# -# b#))
 
 {-# INLINE bytesToBits #-}
 bitsToBytes :: NumBits -> (NumBytes, NumBits)
-bitsToBytes (NumBits n) = (NumBytes $ n `shiftR` 3, NumBits $ n .&. 7)
+bitsToBytes (NumBits (I# n#)) = ( NumBytes (I# (n# `uncheckedIShiftRL#` 3#))
+                                , NumBits (I# (n# `andI#` 7#))
+                                )
 
 {-# INLINE bitsToBytes #-}
 bytesToBits :: NumBytes -> NumBits
-bytesToBits (NumBytes n) = NumBits $ n `shiftL` 3
+bytesToBits (NumBytes (I# n#)) = NumBits (I# (n# `uncheckedIShiftL#` 3#))
 
 ----------------------------------------------------------------------
 
@@ -90,17 +99,19 @@ emptyBitString = BitString (NumBits 0) 0
 -- BitString.
 
 joinBitString :: BitString -> BitString -> BitString
-joinBitString a b =
-  let bitSizeA = bitCount $ bsLength a
-  in BitString { bsLength = addBitCounts (bsLength a) (bsLength b)
-               , bsData = bsData a .|. (bsData b `shiftL` bitSizeA)
-               }
+joinBitString (BitString (Bits' (I# szA#)) (I# a#))
+              (BitString (Bits' (I# szB#)) (I# b#)) =
+  BitString { bsLength = NumBits (I# (szA# +# szB#))
+            , bsData = I# (a# `orI#` (b# `uncheckedIShiftL#` szA#))
+            }
 
 
 -- | Given a number of bits to take, and an @Integer@, create a @BitString@.
 
 toBitString :: NumBits -> Int -> BitString
-toBitString len val = BitString len (val .&. maskBits len)
+toBitString len@(Bits' (I# len#)) (I# val#) =
+  let !mask# = (1# `uncheckedIShiftL#` len#) -# 1#
+  in BitString len (I# (val# `andI#` mask#))
 
 
 -- | Extract the referenced Integer value from a BitString
@@ -140,13 +151,6 @@ showBitString bs = showString padding . showString bin
   fmt _   = error "invalid binary digit value"
 
 
--- | Generate a mask from a number of bits desired.
-maskBits :: NumBits -> Int
-maskBits (NumBits len)
-  | len <= 0  = 0
-  | otherwise = pred (bit len)
-
-
 -- | Extract a smaller BitString with the specified number of bits from the
 -- \"start\" of a larger BitString.
 take :: NumBits -> BitString -> BitString
@@ -159,8 +163,10 @@ take n bs@(BitString l i)
 -- return the remaining as a smaller BitString.
 
 drop :: NumBits -> BitString -> BitString
-drop n (BitString l i)
+drop !n !(BitString l i)
   | n >= l    = emptyBitString
-  | otherwise = BitString
-                (NumBits $ bitCount l - bitCount n)
-                (i `shiftR` (bitCount n))
+  | otherwise =
+      let !(I# n#) = bitCount n
+          !(I# l#) = bitCount l
+          !(I# i#) = i
+      in BitString (NumBits (I# (l# -# n#))) (I# (i# `uncheckedIShiftRL#` n#))
--- a/src/Data/LLVM/BitCode/BitString.hs
+++ b/src/Data/LLVM/BitCode/BitString.hs
@@ -74,13 +74,13 @@ data BitString = BitString
     -- unlimited size value.  However, this adds some overhead to various
     -- computations, and since LLVM Bitcode is unlikely to ever represent values
     -- greater than the native size (64 bits) as discrete values.  By changing
-    -- this to @Int@, the use of uboxed calculations is enabled for better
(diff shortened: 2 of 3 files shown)


src/Data/LLVM/BitCode/GetBits.hs
```diff
@@ -125,112 +117,233 @@ instance MonadPlus GetBits where
...
+          then
+            let !s8# = sBit# `uncheckedIShiftRL#` 3#
+                !hop# = sBit# `andI#` 7#
+                !r8# = ((hop# +# nbits# +# 7#) `uncheckedIShiftRL#` 3#)
+                !mask# = (1# `uncheckedIShiftL#` nbits#) -# 1#
+
+#if MIN_VERSION_base(4,16,0)
+                word8ToInt !w8# = word2Int# (word8ToWord# w8#)
+#else
+-- technically #if !MIN_VERSION_ghc_prim(0,8,0), for GHC 9.2, but that isn't a define
```
RyanGlScott, line 176: More precisely, `MIN_VERSION_ghc_prim` isn't defined because `llvm-pretty-bc-parser` doesn't declare an explicit dependency on `ghc-prim`. But `base`'s `GHC.Exts` does re-export everything of interest from `ghc-prim`, so we could be a little more precise by changing this CPP to `MIN_VERSION_base(4,16,0)`.

src/Data/LLVM/BitCode/GetBits.hs
```diff
@@ -125,112 +117,233 @@ instance MonadPlus GetBits where
...
-      -- excess high bits.
-      vi = wi `shiftR` (bitCount startBit .&. 7) .&. mask
+                      -> Either String (() -> (# Int#, Int# #))
+extractFromByteString !bitLim# !sBit# !nbits# bs =
+     if isTrue# ((1# `uncheckedIShiftL#` (nbits#)) /=# 0#)
+        -- (nbits# -# 1#) above would allow 64-bit value extraction, but this
+        -- function cannot actually support a size of 64, because Int# is signed,
+        -- so it doesn't properly use the high bit in numeric operations.  This
+        -- seems to be OK at this point because LLVM bitcode does not attempt to
+        -- encode actual 64-bit values.
```
RyanGlScott, line 163: I'm not sure what you mean by "LLVM bitcode does not attempt to encode actual 64-bit values". After all, `i64` is most definitely a type in LLVM, which `(u)int64_t` compiles down to.

src/Data/LLVM/BitCode/GetBits.hs
```diff
@@ -125,112 +117,233 @@ instance MonadPlus GetBits where
...
 -- with the bit position following the extraction.
 
-extractFromByteString :: NumBits {-^ the last bit accessible in the ByteString -}
-                      -> NumBits {-^ the bit to start extraction at -}
-                      -> NumBits {-^ the number of bits to extract -}
+-- There are two implementations: one builds an integer value by shifting bits,
+-- then shifts and masks the result to get the final value.  The other uses
+-- unlifted values and avoids the final shift by being smarter about individual
+-- compositions.  Their functionality should be identical, but it may be easier
+-- to debug the first.
```
RyanGlScott, line 124: Perhaps it's worth defining the simpler version under a different name. It wouldn't actually be used anywhere, but that way the code would still be typechecked, making it less susceptible to bitrot.

src/Data/LLVM/BitCode/GetBits.hs
```diff
@@ -26,91 +30,79 @@ import qualified Control.Monad.Fail
 
 -- Bit-level Parsing -----------------------------------------------------------
 
-newtype GetBits a = GetBits { unGetBits :: Either String (BitPosition -> BS.ByteString -> (BitsGetter a, BitPosition)) } -- Left is outer fail
+newtype GetBits a =
+  GetBits { unGetBits :: BitPosition -> BS.ByteString
+                      -> (# BitsGetter a, BitPosition #)
+          }
```
RyanGlScott, line 36: This removes the outer fail by getting rid of the `Either`. Does this mean that the `Either` in `BitsGetter` handles both forms of failure now?

src/Data/LLVM/BitCode/BitString.hs
```diff
@@ -40,27 +44,48 @@ pattern Bytes' n = NumBytes n
...
 ----------------------------------------------------------------------
 
 data BitString = BitString
   { bsLength :: !NumBits
-  , bsData   :: !Integer
+  , bsData   :: !Int
+    -- Note: the bsData was originally an Integer, which allows an essentially
+    -- unlimited size value.  However, this adds some overhead to various
+    -- computations, and since LLVM Bitcode is unlikely to ever represent values
+    -- > native size (64 bits) as discrete values.  By changing this to @Int@,
```
RyanGlScott, line 76: Is the use of `>` in `> native size [...]` here (which Haddock interprets as inline code) a mistake?

src/Data/LLVM/BitCode/BitString.hs
```diff
@@ -40,27 +44,48 @@ pattern Bytes' n = NumBytes n
...
+    -- the use of uboxed calculations is enabled for better performance.
+    --
+    -- The use of Int is potentially unsound because GHC only guarantees it's a
+    -- signed 32-bit integer.  However current implementation in all environments
+    -- where it's reasonable to use this parser have a 64-bit Int
+    -- implementatinon.  This can be verified via:
```
RyanGlScott, line 82:
```suggestion
    -- where it's reasonable to use this parser have a 64-bit Int
    -- implementation.  This can be verified via:
```

src/Data/LLVM/BitCode/GetBits.hs
```diff
@@ -125,112 +117,233 @@ instance MonadPlus GetBits where
...
-      -- Combine the extracted bytes into an Integer value in wi.
-      wi = BS.foldr (\w a -> a `shiftL` 8 .|. fromIntegral w) (0::Integer) ws
-
-      -- Mask is 0-bit based set of bits wanted in the result
-      mask = ((1::Integer) `shiftL` bitCount numBits) - 1
-
-      -- Shift the desired value down to byte alignment and then discard any
-      -- excess high bits.
-      vi = wi `shiftR` (bitCount startBit .&. 7) .&. mask
+                      -> Either String (() -> (# Int#, Int# #))
```
RyanGlScott, line 156: The use of `() -> (# Int#, Int# #)` here is rather awkward. I get the sense that the `()` argument was only added to make it work with `Either`, which expects boxed arguments. Why not use `UnboxedSums` instead? Something like this, to be more precise:

```hs
(# String | (# Int#, Int# #) #)
```

src/Data/LLVM/BitCode/GetBits.hs
kquick: Ahh, I guess I hadn't tracked the missing dependency/re-export issue.  That said, I'm not sure I understand your suggestion... I think that what I have already?

src/Data/LLVM/BitCode/GetBits.hs
```diff
@@ -125,112 +117,233 @@ instance MonadPlus GetBits where
...
+                    _ -> let !(W8# w#) = (bs `BS.index` (I# (s8# +# i#)))
+                         in (word8ToInt w#) `uncheckedIShiftL#` (8# *# i#)
+                getSB# :: Int# -> Int#
+                getSB# !i# =
+                  case i# of
+                    0# -> let !(W8# w#) = bs `BS.index` (I# s8#)
+                          in (word8ToInt w#) `uncheckedIShiftRL#` hop#
+                    _  -> let !(W8# w#) = bs `BS.index` (I# (s8# +# i#))
+                              !shft# = (8# *# i#) -# hop#
+                          in (word8ToInt w#) `uncheckedIShiftL#` shft#
```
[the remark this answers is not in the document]
kquick: Added some comments.

src/Data/LLVM/BitCode/GetBits.hs
```diff
@@ -125,112 +117,233 @@ instance MonadPlus GetBits where
...
+                                       getSB# 8#
+                                 18# -> getSB# 0# `orI#` getSB# 1# `orI#`
+                                        getSB# 2# `orI#` getSB# 3# `orI#`
+                                        getSB# 4# `orI#` getSB# 5# `orI#`
+                                        getSB# 6# `orI#` getSB# 7# `orI#`
+                                        getSB# 8# `orI#` getSB# 9# `orI#`
+                                        getSB# 10# `orI#` getSB# 11# `orI#`
+                                        getSB# 12# `orI#` getSB# 13# `orI#`
+                                        getSB# 14# `orI#` getSB# 15# `orI#`
+                                        getSB# 16# `orI#` getSB# 17#
```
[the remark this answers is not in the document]
kquick: I'm hand-unrolling the `foldr` catch-all case value for sizes that seemed to be used frequently in parsing the bitcode samples I was working with.

src/Data/LLVM/BitCode/GetBits.hs
kquick: This is referring to the actual encoding sizes rather than the native sizes.  These are mostly used for various fields in the AST structure, whereas constant values tend to be encoded as a vbr6, which is capable of larger sizes.  That said, it's not definitive, but I've also never had this fail in the half-dozen different bitcode files I've tested it on.  The comment is a little dated though because I'll be switching to `Word64#` representation in an upcoming PR.

src/Data/LLVM/BitCode/GetBits.hs
kquick: That is indeed the case.  I'll re-address this in a future PR.

src/Data/LLVM/BitCode/GetBits.hs
kquick: Good idea.

src/Data/LLVM/BitCode/GetBits.hs
```diff
@@ -26,91 +30,79 @@ import qualified Control.Monad.Fail
...
 -- Bit-level Parsing -----------------------------------------------------------
 
-newtype GetBits a = GetBits { unGetBits :: Either String (BitPosition -> BS.ByteString -> (BitsGetter a, BitPosition)) } -- Left is outer fail
+newtype GetBits a =
+  GetBits { unGetBits :: BitPosition -> BS.ByteString
+                      -> (# BitsGetter a, BitPosition #)
+          }
 
-newtype BitPosition = BitPosition (NumBits, NumBits) -- (current, limit)
+type BitPosition = (# Int#, Int# #)
```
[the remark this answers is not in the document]
kquick: Good point, re-added.

src/Data/LLVM/BitCode/BitString.hs
```diff
@@ -101,15 +128,17 @@ fromBitString (BitString l i) =
...
       | otherwise -> error (unwords
            [ "Data.LLVM.BitCode.BitString.fromBitString: bitstring value of length", show l
-           , "(", show i, ")"
+           , "(mask=0x" <> showHex i ")"
            , "could not be parsed into type with only", show n, "bits"
            ])
  where
- x    = fromInteger ival
- ival = i .&. maskBits l
+ x    = fromInteger ival  -- use Num's conversion
```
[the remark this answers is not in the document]
kquick: It's not necessarily to an `Int`.  This function is polymorphic in the return type, as long as the return type conforms to `Num`, so it's round-tripping through `Integer` to get to the return type, as mentioned in the haddocks.   I added some additional comments, let me know if you think it could still use clarification.

src/Data/LLVM/BitCode/BitString.hs
```diff
@@ -40,27 +44,48 @@ pattern Bytes' n = NumBytes n
...
+    -- The use of Int is potentially unsound because GHC only guarantees it's a
+    -- signed 32-bit integer.  However current implementation in all environments
+    -- where it's reasonable to use this parser have a 64-bit Int
+    -- implementatinon.  This can be verified via:
+    --
+    --  > import Data.Bits
+    --  > bitSizeMaybe (maxBound :: Int) >= Just 64
+    --
+    -- There's no good location here to automate this check (perhaps
+    -- GetBits.hs:runGetBits?), which is why it isn't verified at runtime.
```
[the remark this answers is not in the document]
kquick: I actually have a compile-time verification (via TH) that will be in an upcoming PR.

src/Data/LLVM/BitCode/BitString.hs
kquick: Fixed.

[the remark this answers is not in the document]
kquick: I was perhaps being a bit dismissive of use cases outside our own.  In a future PR, this will migrate to 64-bit minimum values.

src/Data/LLVM/BitCode/BitString.hs
```diff
@@ -40,27 +44,48 @@ pattern Bytes' n = NumBytes n
...
-  , bsData   :: !Integer
+  , bsData   :: !Int
+    -- Note: the bsData was originally an Integer, which allows an essentially
+    -- unlimited size value.  However, this adds some overhead to various
+    -- computations, and since LLVM Bitcode is unlikely to ever represent values
+    -- > native size (64 bits) as discrete values.  By changing this to @Int@,
+    -- the use of uboxed calculations is enabled for better performance.
+    --
+    -- The use of Int is potentially unsound because GHC only guarantees it's a
+    -- signed 32-bit integer.  However current implementation in all environments
```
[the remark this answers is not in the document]
kquick: Yup, that's what I meant.

src/Data/LLVM/BitCode/BitString.hs
kquick: It was an auto-formatting in the editor than ended up with that as the first character... good catch!

llvm-pretty-bc-parser.cabal
```diff
@@ -72,7 +75,7 @@ Library
 Executable llvm-disasm
   Main-is:             LLVMDis.hs
   Default-language:    Haskell2010
-  Ghc-options:         -Wall
+  Ghc-options:         -Wall -Wunbanged-strict-patterns -O2
```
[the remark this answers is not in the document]
kquick: I don't think the application has any strict fields, but I also didn't really check, and it certainly won't hurt anything, so I've added it.

src/Data/LLVM/BitCode/GetBits.hs
RyanGlScott: Oops, so it is. In that case, I suppose the only thing left to do is to refine the comment a bit.

src/Data/LLVM/BitCode/GetBits.hs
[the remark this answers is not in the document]
RyanGlScott: I see. Please add that as a comment—that definitely wasn't obvious when reading this.

src/Data/LLVM/BitCode/GetBits.hs
RyanGlScott: Ah, I see. Since this will change in an upcoming PR, I'll let this be.

src/Data/LLVM/BitCode/BitString.hs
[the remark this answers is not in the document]
RyanGlScott: Oops, I missed the fact that this was polymorphic. I like the new comments.

RyanGlScott (approved):

PR #198, opened 2022-09-13
Add quick flag and use unsigned unlifted Word#

kquick:
This makes further performance modifications to use unsigned `Word#` instead of signed `Int#` and adds a `quick` flag (disabled by default).  When the `quick` flag is enabled, much of the error detection, verification, and other internal checks are removed; this mode is a further incremental performance gain but it should only be used in situations where well-formed bitcode from known-supported versions of LLVM willl be used.

### DIFF
--- a/src/Data/LLVM/BitCode/BitString.hs
+++ b/src/Data/LLVM/BitCode/BitString.hs
@@ -2,6 +2,8 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE MagicHash #-}
 {-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE TemplateHaskell #-}
 
 module Data.LLVM.BitCode.BitString
   (
@@ -16,6 +18,7 @@ module Data.LLVM.BitCode.BitString
   , NumBits, NumBytes, pattern Bits', pattern Bytes'
   , bitCount, bitCount#
   , bitsToBytes, bytesToBits
+  , bitMask
   , addBitCounts
   , subtractBitCounts
   )
@@ -31,6 +34,7 @@ import Numeric ( showIntAtBase, showHex )
 import GHC.Exts
 
 import Prelude hiding (take,drop,splitAt)
+
 
 ----------------------------------------------------------------------
 -- Define some convenience newtypes to clarify whether the count of bits or count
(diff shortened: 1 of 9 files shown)


src/Data/LLVM/BitCode/BitString.hs
```diff
@@ -100,31 +111,36 @@ emptyBitString = BitString (NumBits 0) 0
...
 bitStringValue = bsData
 
 
 -- | Extract a target (Num) value of the desired type from a BitString (using
 -- fromInteger to perform the target type conversion).
 
 fromBitString :: (Num a, Bits a) => BitString -> a
+#ifdef QUICK
+fromBitString (BitString _ i) = x
+#else
```
RyanGlScott, line 143: I'm not a fan of sprinkling CPP everywhere throughout the library like this, since (1) it makes adds a lot of lexical noise, and (2) it makes it more difficult to remember to keep the two versions of every function in sync, since you have to remember to build with both `-fquick` and `-f-quick` unless you want to risk bitrot.

Rather this going with this approach, why not do something like this instead?

```hs
fromBitString :: (Num a, Bits a) => BitString -> a
fromBitString (BitString l i)
  | quickMode = x
  | otherwise =
    case bitSizeMaybe x of
      ...
```

Where:

```hs
quickMode :: Bool
#if defined(QUICK)
quickMode = True
#else
quickMode = False
#endif
```

This way, we can continue to typecheck both versions of each function, decreasing the chances of bitrot. Since `quickMode` will statically be known to be `True` or `False`, the optimizer will be able to make swift work of it.

src/Data/LLVM/BitCode/GetBits.hs
```diff
@@ -159,144 +159,151 @@ extractFromByteString :: Int# {-^ the last bit accessible in the ByteString -}
                       -> Int# {-^ the bit to start extraction at -}
                       -> Int# {-^ the number of bits to extract -}
                       -> ByteString {-^ the ByteString to extract from -}
-                      -> Either String (() -> (# Int#, Int# #))
+                      -> Either String (() -> (# Word#, Int# #))
 extractFromByteString !bitLim# !sBit# !nbits# bs =
```
RyanGlScott, line 163: Now that we have a `quick` flag, is it worth making `extractFromByteString'` above the `quick` version of this function?

src/Data/LLVM/BitCode/IR/Constants.hs
```diff
@@ -425,10 +425,14 @@ parseConstantEntry t (getTy,cs) (fromEntry -> Just r) =
         _asmDialect    = mask `shiftR` 2
 
     asmStrSize <- field 1 numeric
+#ifndef QUICK
     Assert.recordSizeGreater r (1 + asmStrSize)
+#endif
```
RyanGlScott, line 430: Similarly, rather than surrounding every assertion with CPP, it would be better to redefine each assertion function in terms of `quickMode`. Something like:

```hs
recordSizeGreater x y = Control.Monad.unless quickMode $ ...
```

PR #200, opened 2022-09-27
Support three GHC versions at a time

langston-barrett:
I'm not tied to this particular policy, I just saw that we were running against five GHC versions in CI and it seemed like a bit much.

Looks like this will require a change to the repository settings to not require the GHC versions that are dropped in this PR.


RyanGlScott (approved):

kquick:
Sorry to be a bit late on this: this is generally fine, although I don't think we should be targeting 9.0  since there were known issues in 9.0.  No need to re-open/change this now, just FYI.

PR #202, opened 2023-03-15
Support LLVM 13

### LINKED ISSUE #174: support `METADATA_ARG_LIST`

### LINKED ISSUE #184: Support `CST_CODE_INLINEASM` constant codes added in LLVM 13 and 14
Prior to LLVM 13, there were two constant codes (18 and 23) for inline assembly statements. There are now two more such codes as of LLVM 14:

* Code 28 (now named `CST_CODE_INLINEASM_OLD3`), added in LLVM 13 in https://github.com/llvm/llvm-project/commit/8ec9fd483949ca3b23053effcac226dcc56e7a95
* Code 30 (now named `CST_CODE_INLINEASM`), added in LLVM 14 in https://github.com/llvm/llvm-project/commit/62b1682570b1059e8c6542192159dcde32c13c30

Currently, `llvm-pretty-bc-parser` does not support either of these. Here is a test case:

```c
int main() {
  asm("nop");
  return 0;
}
```

When compiled with `~/Software/clang+llvm-13.0.1/bin/clang test.c -c -emit-llvm`, this produces the following parse error:

```
λ> parseBitCodeFromFile "test.bc"
Left (Error {errContext = ["CONSTANTS_BLOCK","@main","FUNCTION_BLOCK","FUNCTION_BLOCK_ID","value symbol table","MODULE_BLOCK","Bitstream"], errMessage = "Unknown constant record code 28\nAre you sure you're using a supported compiler?\nCheck here: https://github.com/GaloisInc/llvm-pretty-bc-parser\n"})
```

The error is much the same when compiled with LLVM 14, except that the code will be 30 instead of 28.

With some effort, we should be able to adapt the existing `llvm-pretty-bc-parser` code for code 18 (now named `CST_CODE_INLINEASM_OLD` as of LLVM 14):

https://github.com/GaloisInc/llvm-pretty-bc-parser/blob/daffc926d526c51ba7967865ec4b713aafd0941f/src/Data/LLVM/BitCode/IR/Constants.hs#L360-L373

And for code 23 (now named `CST_CODE_INLINEASM_OLD2` as of LLVM 14):

https://github.com/GaloisInc/llvm-pretty-bc-parser/blob/daffc926d526c51ba7967865ec4b713aafd0941f/src/Data/LLVM/BitCode/IR/Constants.hs#L416-L437

RyanGlScott:
This contains two key fixes needed to support LLVM 13:

## Support `METADATA_ARG_LIST`

Fixes #174.

## Support newer inline `asm` codes

One code (`CST_CODE_INLINEASM_OLD3`) was introduced in LLVM 13, and another (`CST_CODE_INLINEASM`) was introduced in LLVM 14. For the most part, they are parsed identically to previous inline `asm` codes, but with some minor differences. I have consolidated the logic for parsing all inline `asm` codes into a single `parseInlineAsm` function.

Fixes #184.

-----

The rest of the commits are needed to make the test case for #174 work, which exercises quite a bit of pretty-printing functionality that was not previously tested. These include fixes for https://github.com/elliottt/llvm-pretty/issues/105, https://github.com/elliottt/llvm-pretty/issues/107, #211, and #212. This also requires bumping the `llvm-pretty` submodule to bring in the changes from https://github.com/elliottt/llvm-pretty/pull/108.

### DIFF
--- a/disasm-test/tests/di-arg-list.at-least-llvm13.ll
+++ b/disasm-test/tests/di-arg-list.at-least-llvm13.ll
@@ -1,0 +1,43 @@
+; ModuleID = 'di-arg-list.c'
+source_filename = "di-arg-list.c"
+target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
+target triple = "x86_64-unknown-linux-gnu"
+
+; Function Attrs: mustprogress nofree norecurse nosync nounwind readnone uwtable willreturn
+define dso_local i32 @f(i32 %0, i32 %1) local_unnamed_addr #0 !dbg !9 {
+  call void @llvm.dbg.value(metadata i32 %0, metadata !14, metadata !DIExpression()), !dbg !17
+  call void @llvm.dbg.value(metadata i32 %1, metadata !15, metadata !DIExpression()), !dbg !17
+  call void @llvm.dbg.value(metadata !DIArgList(i32 %0, i32 %1), metadata !16, metadata !DIExpression(DW_OP_LLVM_arg, 0, DW_OP_LLVM_arg, 1, DW_OP_plus, DW_OP_stack_value)), !dbg !17
+  ret i32 0, !dbg !18
+}
+
+; Function Attrs: nofree nosync nounwind readnone speculatable willreturn
+declare void @llvm.dbg.value(metadata, metadata, metadata) #1
+
+attributes #0 = { mustprogress nofree norecurse nosync nounwind readnone uwtable willreturn "frame-pointer"="none" "min-legal-vector-width"="0" "no-trapping-math"="true" "stack-protector-buffer-size"="8" "target-cpu"="x86-64" "target-features"="+cx8,+fxsr,+mmx,+sse,+sse2,+x87" "tune-cpu"="generic" }
+attributes #1 = { nofree nosync nounwind readnone speculatable willreturn }
+
+!llvm.dbg.cu = !{!0}
+!llvm.module.flags = !{!3, !4, !5, !6}
+!llvm.ident = !{!7}
+!llvm.commandline = !{!8}
+
+!0 = distinct !DICompileUnit(language: DW_LANG_C99, file: !1, producer: "clang version 13.0.1", isOptimized: true, runtimeVersion: 0, emissionKind: FullDebug, enums: !2, splitDebugInlining: false, nameTableKind: None)
+!1 = !DIFile(filename: "di-arg-list.c", directory: "/home/rscott/Documents/Hacking/Haskell/llvm-pretty-bc-parser/disasm-test/tests")
+!2 = !{}
+!3 = !{i32 7, !"Dwarf Version", i32 4}
+!4 = !{i32 2, !"Debug Info Version", i32 3}
+!5 = !{i32 1, !"wchar_size", i32 4}
+!6 = !{i32 7, !"uwtable", i32 1}
+!7 = !{!"clang version 13.0.1"}
+!8 = !{!"/home/rscott/Software/clang+llvm-13.0.1/bin/clang-13 -S -emit-llvm -g -O1 -frecord-command-line di-arg-list.c -o di-arg-list.at-least-llvm13.ll"}
+!9 = distinct !DISubprogram(name: "f", scope: !1, file: !1, line: 1, type: !10, scopeLine: 1, flags: DIFlagPrototyped | DIFlagAllCallsDescribed, spFlags: DISPFlagDefinition | DISPFlagOptimized, unit: !0, retainedNodes: !13)
+!10 = !DISubroutineType(types: !11)
+!11 = !{!12, !12, !12}
+!12 = !DIBasicType(name: "int", size: 32, encoding: DW_ATE_signed)
+!13 = !{!14, !15, !16}
+!14 = !DILocalVariable(name: "x", arg: 1, scope: !9, file: !1, line: 1, type: !12)
+!15 = !DILocalVariable(name: "y", arg: 2, scope: !9, file: !1, line: 1, type: !12)
+!16 = !DILocalVariable(name: "z", scope: !9, file: !1, line: 2, type: !12)
+!17 = !DILocation(line: 0, scope: !9)
+!18 = !DILocation(line: 3, column: 3, scope: !9)
--- a/src/Data/LLVM/BitCode/IR/Constants.hs
+++ b/src/Data/LLVM/BitCode/IR/Constants.hs
@@ -360,19 +360,8 @@ parseConstantEntry t (getTy,cs) (fromEntry -> Just r) =
     return (getTy, Typed (PrimType (Integer 1)) (ValConstExpr cst):cs)
 
   18 -> label "CST_CODE_INLINEASM_OLD" $ do
-    let field = parseField r
-    ty    <- getTy
-    flags <- field 0 numeric
-    let sideEffect = testBit (flags :: Int) 0
-        alignStack = (flags `shiftR` 1) == 1
-
-    alen <- field 1 numeric
-    asm  <- UTF8.decode <$> parseSlice r 2 alen char
-
-    clen <- field (2+alen) numeric
-    cst  <- UTF8.decode <$> parseSlice r (3+alen) clen char
-
-    return (getTy, Typed ty (ValAsm sideEffect alignStack asm cst):cs)
+    tv <- parseInlineAsm InlineAsmCode18 getTy r
+    return (getTy, tv:cs)
 
   19 -> label "CST_CODE_CE_SHUFFLEVEC_EX" $ do
     notImplemented
@@ -415,28 +404,9 @@ parseConstantEntry t (getTy,cs) (fromEntry -> Just r) =
       FloatType Double   -> build (ValDouble . castDouble)
       x                  -> Assert.unknownEntity "element type" x
 
-  23 -> label "CST_CODE_INLINEASM" $ do
-    let field = parseField r
-    mask <- field 0 numeric
-
-    let test = testBit (mask :: Word32)
-        hasSideEffects = test 0
-        isAlignStack   = test 1
-        _asmDialect    = mask `shiftR` 2
-
-    asmStrSize <- field 1 numeric
-    Assert.recordSizeGreater r (1 + asmStrSize)
-
-    constStrSize <- field (2 + asmStrSize) numeric
-    Assert.recordSizeGreater r (2 + asmStrSize + constStrSize)
-
-    asmStr   <- fmap UTF8.decode $ parseSlice r  2               asmStrSize   char
-    constStr <- fmap UTF8.decode $ parseSlice r (3 + asmStrSize) constStrSize char
-
-    ty <- getTy
-    let val = ValAsm hasSideEffects isAlignStack asmStr constStr
-
-    return (getTy, Typed ty val : cs)
+  23 -> label "CST_CODE_INLINEASM_OLD2" $ do
+    tv <- parseInlineAsm InlineAsmCode23 getTy r
+    return (getTy, tv:cs)
 
   -- [opty, flags, n x operands]
   24 -> label "CST_CODE_CE_GEP_WITH_INRANGE_INDEX" $ do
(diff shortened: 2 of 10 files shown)


src/Data/LLVM/BitCode/IR/Constants.hs
```diff
@@ -461,6 +431,20 @@ parseConstantEntry t (getTy,cs) (fromEntry -> Just r) =
     ty <- getTy
     return (getTy, Typed ty ValPoison : cs)
 
+  27 -> label "CST_CODE_DSO_LOCAL_EQUIVALENT" $ do
+    notImplemented
```
langston-barrett, line 435: (I know this PR is still a draft, but): Should we make an issue for this, and the other `notImplemented` below?

RyanGlScott: Yes, certainly. I'll do this in due course.

langston-barrett: Great, thank you!

RyanGlScott: I've opened #208 and #209 for these.

RyanGlScott:
I have good news and I have bad news.

The good news is that I have merged #210, which will allow us to write test cases that require LLVM 13 or later.

The bad news is that while writing a test case for #174, I discovered a slew of metadata-related bugs in `llvm-pretty` and `llvm-pretty-bc-parser`:

* https://github.com/elliottt/llvm-pretty/issues/105
* https://github.com/elliottt/llvm-pretty/issues/107
* https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/211

None of these issues are specifically about LLVM 13—they're quite old bugs that have laid dormant due to `disasm-test` not having any substantial test cases involving metadata. I expect that it will be straightforward to fix these bugs, although I might have to make this PR a bit more bloated in order to get all of the prerequisite bugfixes in place.

RyanGlScott:
It took me longer than I'd hoped, but I've finally managed to add a regression test for #174—see `di-arg-list.at-least-llvm13.ll`. Thankfully, the existing `callbr.ll` test case already works as a regression test for #184.

In order to make all of the metadata in `di-arg-list.at-least-llvm13.ll` survive a roundtrip through `llvm-pretty`'s pretty-printer, I had to fix quite a few unrelated bugs as preparation. As a result, this PR is a bit more bloated than I'd like, but oh well. I've made an effort to put each change into its own commit, so that is the best way to review this patch.

-----

@kquick, I'm experiencing an issue with the `tasty-sugar` setup we're using, and I'm wondering if you have any ideas on how to resolve it. As we already have an `at-least-llvm12` option in `validParams`, I went ahead and added an `at-least-llvm13` option in this pull request as well. This has an unfortunate side effect when running `disasm-test` with LLVM 13, however: it doubles the number of tests! Compare when happens when I run the test suite with LLVM 12:

```
$ PATH=~/Software/clang+llvm-12.0.1/bin:$PATH cabal run disasm-test -fregression -ffuzz -- -l
Disassembly tests.llvm-as 12.0.1.anon-forward-ref.ll.at-least-llvm12.anon-forward-ref
Disassembly tests.llvm-as 12.0.1.atomicrmw.ll.at-least-llvm12.atomicrmw
Disassembly tests.llvm-as 12.0.1.callbr.ll.at-least-llvm12.callbr
Disassembly tests.llvm-as 12.0.1.factorial2.ll.at-least-llvm12.factorial2
Disassembly tests.llvm-as 12.0.1.fcmp-fast-math.ll.at-least-llvm12.fcmp-fast-math
Disassembly tests.llvm-as 12.0.1.fn-metadata.ll.at-least-llvm12.fn-metadata
Disassembly tests.llvm-as 12.0.1.fneg.ll.at-least-llvm12.fneg
Disassembly tests.llvm-as 12.0.1.fun-attrs.ll.at-least-llvm12.fun-attrs
Disassembly tests.llvm-as 12.0.1.global-array.ll.at-least-llvm12.global-array
Disassembly tests.llvm-as 12.0.1.half-float.ll.at-least-llvm12.half-float
Disassembly tests.llvm-as 12.0.1.hello-world.ll.at-least-llvm12.hello-world
Disassembly tests.llvm-as 12.0.1.insertelt.ll.at-least-llvm12.insertelt
Disassembly tests.llvm-as 12.0.1.mutual-struct-rec.ll.at-least-llvm12.mutual-struct-rec
Disassembly tests.llvm-as 12.0.1.phi-anon.ll.at-least-llvm12.phi-anon
Disassembly tests.llvm-as 12.0.1.poison.at-least-llvm12.ll.at-least-llvm12.poison
Disassembly tests.llvm-as 12.0.1.simple-metadata.ll.at-least-llvm12.simple-metadata
Disassembly tests.llvm-as 12.0.1.simple-res.ll.at-least-llvm12.simple-res
Disassembly tests.llvm-as 12.0.1.struct-initializer.ll.at-least-llvm12.struct-initializer
Disassembly tests.llvm-as 12.0.1.switch-big-value.ll.at-least-llvm12.switch-big-value
Disassembly tests.llvm-as 12.0.1.switch-same-target.ll.at-least-llvm12.switch-same-target
Disassembly tests.llvm-as 12.0.1.switch-simple.ll.at-least-llvm12.switch-simple
Disassembly tests.llvm-as 12.0.1.switch.ll.at-least-llvm12.switch
```

Versus LLVM 13:

```
$ PATH=~/Software/clang+llvm-13.0.0/bin:$PATH cabal run disasm-test -fregression -ffuzz -- -l
Disassembly tests.llvm-as 13.0.0.anon-forward-ref.ll.at-least-llvm12.anon-forward-ref
Disassembly tests.llvm-as 13.0.0.anon-forward-ref.ll.at-least-llvm13.anon-forward-ref
Disassembly tests.llvm-as 13.0.0.atomicrmw.ll.at-least-llvm12.atomicrmw
Disassembly tests.llvm-as 13.0.0.atomicrmw.ll.at-least-llvm13.atomicrmw
Disassembly tests.llvm-as 13.0.0.callbr.ll.at-least-llvm12.callbr
Disassembly tests.llvm-as 13.0.0.callbr.ll.at-least-llvm13.callbr
Disassembly tests.llvm-as 13.0.0.di-arg-list.at-least-llvm13.ll.at-least-llvm13.di-arg-list
Disassembly tests.llvm-as 13.0.0.factorial2.ll.at-least-llvm12.factorial2
Disassembly tests.llvm-as 13.0.0.factorial2.ll.at-least-llvm13.factorial2
Disassembly tests.llvm-as 13.0.0.fcmp-fast-math.ll.at-least-llvm12.fcmp-fast-math
Disassembly tests.llvm-as 13.0.0.fcmp-fast-math.ll.at-least-llvm13.fcmp-fast-math
Disassembly tests.llvm-as 13.0.0.fn-metadata.ll.at-least-llvm12.fn-metadata
Disassembly tests.llvm-as 13.0.0.fn-metadata.ll.at-least-llvm13.fn-metadata
Disassembly tests.llvm-as 13.0.0.fneg.ll.at-least-llvm12.fneg
Disassembly tests.llvm-as 13.0.0.fneg.ll.at-least-llvm13.fneg
Disassembly tests.llvm-as 13.0.0.fun-attrs.ll.at-least-llvm12.fun-attrs
Disassembly tests.llvm-as 13.0.0.fun-attrs.ll.at-least-llvm13.fun-attrs
Disassembly tests.llvm-as 13.0.0.global-array.ll.at-least-llvm12.global-array
Disassembly tests.llvm-as 13.0.0.global-array.ll.at-least-llvm13.global-array
Disassembly tests.llvm-as 13.0.0.half-float.ll.at-least-llvm12.half-float
Disassembly tests.llvm-as 13.0.0.half-float.ll.at-least-llvm13.half-float
Disassembly tests.llvm-as 13.0.0.hello-world.ll.at-least-llvm12.hello-world
Disassembly tests.llvm-as 13.0.0.hello-world.ll.at-least-llvm13.hello-world
Disassembly tests.llvm-as 13.0.0.insertelt.ll.at-least-llvm12.insertelt
Disassembly tests.llvm-as 13.0.0.insertelt.ll.at-least-llvm13.insertelt
Disassembly tests.llvm-as 13.0.0.mutual-struct-rec.ll.at-least-llvm12.mutual-struct-rec
Disassembly tests.llvm-as 13.0.0.mutual-struct-rec.ll.at-least-llvm13.mutual-struct-rec
Disassembly tests.llvm-as 13.0.0.phi-anon.ll.at-least-llvm12.phi-anon
Disassembly tests.llvm-as 13.0.0.phi-anon.ll.at-least-llvm13.phi-anon
Disassembly tests.llvm-as 13.0.0.poison.at-least-llvm12.ll.at-least-llvm12.poison
Disassembly tests.llvm-as 13.0.0.simple-metadata.ll.at-least-llvm12.simple-metadata
Disassembly tests.llvm-as 13.0.0.simple-metadata.ll.at-least-llvm13.simple-metadata
Disassembly tests.llvm-as 13.0.0.simple-res.ll.at-least-llvm12.simple-res
Disassembly tests.llvm-as 13.0.0.simple-res.ll.at-least-llvm13.simple-res
Disassembly tests.llvm-as 13.0.0.struct-initializer.ll.at-least-llvm12.struct-initializer
Disassembly tests.llvm-as 13.0.0.struct-initializer.ll.at-least-llvm13.struct-initializer
Disassembly tests.llvm-as 13.0.0.switch-big-value.ll.at-least-llvm12.switch-big-value
Disassembly tests.llvm-as 13.0.0.switch-big-value.ll.at-least-llvm13.switch-big-value
Disassembly tests.llvm-as 13.0.0.switch-same-target.ll.at-least-llvm12.switch-same-target
Disassembly tests.llvm-as 13.0.0.switch-same-target.ll.at-least-llvm13.switch-same-target
Disassembly tests.llvm-as 13.0.0.switch-simple.ll.at-least-llvm12.switch-simple
Disassembly tests.llvm-as 13.0.0.switch-simple.ll.at-least-llvm13.switch-simple
Disassembly tests.llvm-as 13.0.0.switch.ll.at-least-llvm12.switch
Disassembly tests.llvm-as 13.0.0.switch.ll.at-least-llvm13.switch
```

In the latter, we run `at-least-llvm12` _and_ `at-least-llvm13` configurations for every test case. Ideally, we'd only run the `at-least-llvm13` ones, however. Is there a way to configure `tasty-sugar` to make this possible?

kquick:
> In the latter, we run `at-least-llvm12` _and_ `at-least-llvm13` configurations for every test case. Ideally, we'd only run the `at-least-llvm13` ones, however. Is there a way to configure `tasty-sugar` to make this possible?


This isn't a tasty-sugar issue per-se, but the `Explicit` v.s. `Assumed` handling in disasm-test (actually, the issue is mostly the latter).

For background, if you have tasty-sugar parameters, those parameters can be present in a filename (e.g. `blah-at-least-llvm12.ll` or missing; in the latter case, if there is a matching *base* filename (e.g. `blah.ll`) then tasty-sugar will offer that as the expected file and indicate that `at-least-llvm12` is assumed.  It will generate one of these Assumed for every parameter value that is not explicitly matched but for which there is a base match.

Thus, we didn't duplicated *every* file: there was only the single match for `poison.at-least-llvm12.ll` because there is no `poison.ll` and the same goes for the new `di-arg-list.at-least-llvm13.ll`.  We did get the Explicit match for those, and did the right thing.

So the main issue is in disasm-test/Main.hs where it handles the `Assumed` case (https://github.com/GaloisInc/llvm-pretty-bc-parser/blob/master/disasm-test/Main.hs#L307).  We don't want it to return `True` for every assumed parameter... only for one of them.  The solution here is to call `specMatchesInstalled` only for the lowest "at-least" value (because `specMatchesInstalled` will properly filter against the actual version number for an `at-least-llvm*` parameter).  Thus, you should be able to replace that line 307 with something like:

```
    Just (TS.Assumed v) | v == "at-least-llvm12" -> specMatchesInstalled v
    Just (TS.Assumed _) -> False
```

RyanGlScott:
Thanks, that explanation makes sense. I had to tweak your original suggestion to make it worked when the assumed parameter is `pre-llvm12`:

```diff
diff --git a/disasm-test/Main.hs b/disasm-test/Main.hs
index faa1dce..0404046 100644
--- a/disasm-test/Main.hs
+++ b/disasm-test/Main.hs
@@ -307,7 +307,11 @@ runTest llvmVer sweet expct
                    ]
           in case lookup "llvm-range" (TS.expParamsMatch expct) of
                Just (TS.Explicit v) -> specMatchesInstalled v
-               Just (TS.Assumed  v) -> specMatchesInstalled v
+               Just (TS.Assumed v)
+                 |  v == "pre-llvm11" || v == "at-least-llvm12"
+                 -> specMatchesInstalled v
+                 |  otherwise
+                 -> False
                _ -> error "llvm-range unknown"
 
 -- | Assemble some llvm assembly, producing a bitcode file in /tmp.
```

But after doing that, everything works as I would expect on LLVM 11, 12, and 13. Nice! And if I understand correctly, this approach is future-proof in that I shouldn't need to modify the `Assumed` logic every time we add another `at-least-llvmXX` version.

-----

I've pushed the fix above, which means that this PR is finally ready for review!

kquick:
> And if I understand correctly, this approach is future-proof

It should be!

disasm-test/tests/di-arg-list.at-least-llvm13.ll
```diff
@@ -0,0 +1,43 @@
+; ModuleID = 'di-arg-list.c'
```
kquick, line 1: Perhaps we should be adding the .c sources for these .ll to the repository as well?

kquick (approved):
That's 13 done... just 14, 15, and 16 to go.

disasm-test/tests/di-arg-list.at-least-llvm13.ll
RyanGlScott: Oops, I meant to check this in but forgot to do so. Fixed.

PR #205, opened 2023-03-16
CI: Use Ubuntu, pin LLVM and OS versions

### LINKED ISSUE #203: CI: Run `llvm-quick-fuzz` on Ubuntu instead of macOS
macOS runners have limited minutes with GitHub Actions and take longer to spin up in general. I don't think there is any fundamental reason why the `llvm-quick-fuzz` job needs to run on macOS, so we should move them over to Ubuntu runners instead.

### LINKED ISSUE #204: CI: Pin the Ubuntu and LLVM versions
Currently, we do not specify exactly what versions of Ubuntu and LLVM we use in the `build.yml` CI job. This means that GitHub Actions is within its rights to silently upgrade the version of LLVM that is used for testing `llvm-pretty-bc-parser`, and indeed, that is the cause of the breakage [here](https://github.com/GaloisInc/llvm-pretty-bc-parser/actions/runs/4430944786/jobs/7773318547?pr=202). We should pin the Ubuntu/LLVM versions to avoid accidental breakage like this in the future.

RyanGlScott:
Fixes #203. Fixes #204.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,8 +38,8 @@ jobs:
         with:
           ghc-version: ${{ matrix.ghc }}
 
-      - uses: actions/cache@v2
-        name: Cache cabal store
+      - uses: actions/cache/restore@v3
+        name: Restore cabal store cache
         with:
           path: |
             ${{ steps.setup-haskell.outputs.cabal-store }}
@@ -58,3 +58,12 @@ jobs:
 
       - name: Haddock
         run: cabal haddock
+
+      - uses: actions/cache/save@v3
+        name: Save cabal store cache
+        if: always()
+        with:
+          path: |
+            ${{ steps.setup-haskell.outputs.cabal-store }}
+            dist-newstyle
+          key: cabal-${{ runner.os }}-${{ matrix.ghc }}-${{ github.sha }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,6 +58,9 @@ jobs:
           key: ${{ env.CACHE_VERSION }}-cabal-${{ matrix.os }}-${{ matrix.ghc }}-${{ github.sha }}
           restore-keys: ${{ env.CACHE_VERSION }}-cabal-${{ matrix.os }}-${{ matrix.ghc }}-
 
+      - name: Configure
+        run: cabal configure --enable-tests -j2 -ffuzz -fregressions
+
       - name: Build
         run: cabal build --enable-tests
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,13 @@ on:
   pull_request:
 
 env:
+  # The CACHE_VERSION can be updated to force the use of a new cache if
+  # the current cache contents become corrupted/invalid.  This can
+  # sometimes happen when (for example) the OS version is changed but
+  # older .so files are cached, which can have various effects
+  # (e.g. cabal complains it can't find a valid version of the "happy"
+  # tool).
+  CACHE_VERSION: 1
   LLVM_VERSION: 12
 
 jobs:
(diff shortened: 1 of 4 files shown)


RyanGlScott:
Successful `llvm-quick-fuzz` run from this PR: https://github.com/GaloisInc/llvm-pretty-bc-parser/actions/runs/4439811385

.github/workflows/llvm-quick-fuzz.yml
```diff
@@ -62,7 +80,16 @@ jobs:
       - shell: bash
         name: Run fuzzing
         run: |
-          cabal exec -ffuzz fuzz-llvm-disasm -- -n 50 --junit-xml results.xml -o results --collapse -I /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include --clang-flags="-O -w" --clang-flags="-O -w -g" -c clangy
+          cabal exec -ffuzz fuzz-llvm-disasm -- -n 50 --junit-xml results.xml -o results --collapse --clang-flags="-O -w" --clang-flags="-O -w -g" -c clangy
```
kquick, line 83: Not something you changed, but any idea why `--clang-flags` is duplicated/updated here?

.github/workflows/build.yml
```diff
@@ -21,21 +31,28 @@ jobs:
           submodules: true
 
       - name: Install LLVM (for running tests)
-        run: sudo apt-get -y install llvm
+        run: |
+          sudo apt-get -y install llvm-${{ env.LLVM_VERSION }}
+          # Inspired by https://gist.github.com/junkdog/70231d6953592cd6f27def59fe19e50d
+          sudo update-alternatives \
+            --install /usr/bin/llvm-config       llvm-config      /usr/bin/llvm-config-${{ env.LLVM_VERSION }} 100 \
```
kquick, line 38: The `install` with priority 100 assumes that it's a higher priority than anything currently used, so that the `auto` mode will choose this llvm.  It might be better to follow this with a `sudo update-alternatives --set llvm-config /usr/bin/llvm-config-${{ env.LLVM_VERSION }}` (I think that's the right syntax... maybe use `update-alternatives --list llvm-config` to verify) to ensure this is the selected alternative.

kquick (approved):
Mostly looks good/straightforward, and using linux for fuzz testing is probably more "canonical" and less prone to the whims of Apple.  Just one suggestion regarding the update-alternatives.

.github/workflows/build.yml
RyanGlScott: I think `--set` only makes sense when you have specified multiple alternatives to choose from. When I tried your suggestion in CI, [it fails with](https://github.com/GaloisInc/llvm-pretty-bc-parser/actions/runs/4440707184/jobs/7794849944?pr=205#step:3:25):

```
update-alternatives: error: no alternatives for llvm-as
```

At the very least, we can run `update-alternatives --list` on each tool to make sure that what we have `--install`ed is what we intended.

RyanGlScott: Oh wait, this only applies to `llvm-config` itself, not anything listed after `--slave`...

.github/workflows/llvm-quick-fuzz.yml
RyanGlScott: Wow. Wooooooow. Removing the (seemingly) duplicate flags _does_ [cause it to fail](https://github.com/GaloisInc/llvm-pretty-bc-parser/actions/runs/4440936039). What's more, the output does _not_ make it obvious what goes wrong. Sigh.

Something smells about this, but this PR has already dragged on long enough. I'll just continue to stick with the status quo of `--clang-flags="-O -w" --clang-flags="-O -w -g"` for now.

kquick: Yikes!  I agree: we should probably have a better syntax/requirement spec here, but it definitely seems involved and beyond the scope of this PR.  Thanks for checking though!

Issue #177, opened 2021-10-27
Support opaque pointers


RyanGlScott:
LLVM is in the process of migrating all of its pointer types from `i8*`, `i32*`, etc. to an opaque `ptr` type, as described [here](https://llvm.org/docs/OpaquePointers.html). LLVM 13 takes an important step in that direction, as it is the first LLVM release to include `ptr` in the LLVM AST. See https://github.com/llvm/llvm-project/commit/2155dc51d700c9fb5f29d79eaacf5e1470e4d8ca.

Note, however, that ordinary `clang` invocations do not yet produce LLVM ASTs that use `ptr` yet. Because `ptr` is still an experimental feature, one must explicitly opt in to it. Here is one way to do this:

```c
// wat.c
void f(int *x) {
  *x += 1;
}
```
```
$ ~/Software/clang+llvm-13.0.0-x86_64-linux-gnu-ubuntu-20.04/bin/clang -c -emit-llvm wat.c -o wat.bc
$ ~/Software/clang+llvm-13.0.0-x86_64-linux-gnu-ubuntu-20.04/bin/llvm-dis --force-opaque-pointers wat.bc -o wat.ll
$ cat wat.ll
; ModuleID = 'wat.bc'
source_filename = "wat.c"
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
target triple = "x86_64-unknown-linux-gnu"

; Function Attrs: noinline nounwind optnone uwtable
define dso_local void @f(ptr %0) #0 {
  %2 = alloca ptr, align 8
  store ptr %0, ptr %2, align 8
  %3 = load ptr, ptr %2, align 8
  %4 = load i32, ptr %3, align 4
  %5 = add nsw i32 %4, 1
  store i32 %5, ptr %3, align 4
  ret void
}

attributes #0 = { noinline nounwind optnone uwtable "frame-pointer"="all" "min-legal-vector-width"="0" "no-trapping-math"="true" "stack-protector-buffer-size"="8" "target-cpu"="x86-64" "target-features"="+cx8,+fxsr,+mmx,+sse,+sse2,+x87" "tune-cpu"="generic" }

!llvm.module.flags = !{!0, !1, !2}
!llvm.ident = !{!3}

!0 = !{i32 1, !"wchar_size", i32 4}
!1 = !{i32 7, !"uwtable", i32 1}
[...]

PR #207, opened 2023-03-20
Give `call`/`callbr` function types instead of pointer types

### LINKED ISSUE #189: `call`/`callbr`'s function type should not be a pointer type
If you parse an `invoke` instruction, such as these one found in `crux-llvm`'s [`invoke-test.ll` test case](https://github.com/GaloisInc/crucible/blob/7d87bdbb651aa59417ea61096555d3bf38c69d50/crux-llvm/test-data/golden/golden/invoke-test.ll#L169-L170), the resulting IR will be:

```hs
...
  (Invoke
     (FunTy
        (PrimType (Integer 32))
        [ PtrTo (PtrTo (PrimType (Integer 8))) ]
        False)
     (ValSymbol
        (Symbol
           "_ZN3std2rt10lang_start28_$u7b$$u7b$closure$u7d$$u7d$17h6e1c450892b1f754E"))
     [ Typed
         { typedType = PtrTo (PtrTo (PrimType (Integer 8)))
         , typedValue = ValIdent (Ident "arg0")
         }
     ]
     (Named (Ident "bb1"))
     (Named (Ident "cleanup")))
...
```

The thing that is peculiar about this IR is that the argument type is a raw `FunTy`. Reading the [LLVM Language Reference Manual entry for `invoke`](https://releases.llvm.org/14.0.0/docs/LangRef.html#invoke-instruction), however, suggests that this should instead be a `PtrTo`:

> Syntax:
> 
> ```
> <result> = invoke [cconv] [ret attrs] [addrspace(<num>)] <ty>|<fnty> <fnptrval>(<function args>) [fn attrs]
>               [operand bundles] to label <normal label> unwind label <exception label>
> 
> ```
> 
> ...
> 
> Arguments:
> 
> This instruction requires several arguments:
> 
> ...
> 
> 6. `fnptrval`: An LLVM value containing a pointer to a function to be invoked. In most cases, this is a direct function invocation, but indirect invoke’s are just as possible, calling an arbitrary pointer to function value.
> 
> ...

Making this a `PtrTo` would also match the treatment for `call` and `callbr`, two closely related instructions. In fact, in the source of of LLVM, the classes for `call`, `callbr`, and `invoke` are all subclasses of the same [`CallBase` class](https://github.com/llvm/llvm-project/blob/50f846d63443781a5c00ec7022c381c4f09f543d/llvm/include/llvm/IR/InstrTypes.h#L1174), so they all store their function type in the same way.

RyanGlScott:
See `Note [Typing function applications]` for the rationale.

This fixes #189. This is also an important step towards being able to support opaque pointers, as described in #177.

### DIFF
--- a/src/Data/LLVM/BitCode/IR/Function.hs
+++ b/src/Data/LLVM/BitCode/IR/Function.hs
@@ -551,6 +551,8 @@ parseFunctionBlockEntry _ t d (fromEntry -> Just r) = case recordCode r of
         `mplus` fail "invalid INVOKE record"
 
     args        <- parseInvokeArgs t va r ix' as
+    -- Use `fty` instead of `typedType f` as the function type, as `typedType f`
+    -- will be a pointer type. See Note [Typing function applications].
     result ret (Invoke fty (typedValue f) args normal unwind) d
 
   14 -> label "FUNC_CODE_INST_UNWIND" (effect Unwind d)
@@ -728,7 +730,9 @@ parseFunctionBlockEntry _ t d (fromEntry -> Just r) = case recordCode r of
     label (show fn) $ do
       (ret,as,va) <- elimFunTy fnty `mplus` fail "invalid CALL record"
       args <- parseCallArgs t va r ix2 as
-      result ret (Call False opTy fn args) d
+      -- Use `fnty` instead of `opTy` as the function type, as `opTy` will be
+      -- a pointer type. See Note [Typing function applications].
+      result ret (Call False fnty fn args) d
 
   -- [Line,Col,ScopeVal, IAVal]
   35 -> label "FUNC_CODE_DEBUG_LOC" $ do
@@ -1003,7 +1007,9 @@ parseFunctionBlockEntry _ t d (fromEntry -> Just r) = case recordCode r of
     label (show fn) $ do
       (ret,as,va) <- elimFunTy fnty `mplus` fail "invalid CALLBR record"
       args <- parseCallArgs t va r ix2 as
-      result ret (CallBr opTy fn args normal indirectDests) d
+      -- Use `fnty` instead of `opTy` as the function type, as `opTy` will be
+      -- a pointer type. See Note [Typing function applications].
+      result ret (CallBr fnty fn args normal indirectDests) d
 
   -- [opty, opval]
   58 -> label "FUNC_CODE_INST_FREEZE" $ do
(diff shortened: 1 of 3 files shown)


kquick (commented):
I'm aligned with the changes, but is there a way to add a test to at least validate the parse/print round-trip is valid (and agrees with `llvm-dis`)?

RyanGlScott:
I'm glad you prompted me to look, because it turns out that `llvm-pretty` is making the same pointer-type/function-type confusion [here](https://github.com/elliottt/llvm-pretty/blob/a272e380bbd44d6af5afba63a70772b2dab6a2d0/src/Text/LLVM/PP.hs#L695-L697) in its pretty-printer:

```hs
ppCallSym :: LLVM => Type -> Value -> Doc
ppCallSym (PtrTo (FunTy res args va))   val        = ppType res <+> ppArgList va (map ppType args) <+> ppValue val
ppCallSym ty              val                      = ppType ty  <+> ppValue val
```

We shouldn't be using `PtrTo` here. I will need to submit an `llvm-pretty` PR to address this.

src/Data/LLVM/BitCode/IR/Function.hs
```diff
@@ -1454,3 +1460,34 @@ getDecodedAtomicRWOp 8  = pure AtomicMin
...
+
+    ...
+
+    * ‘fnty’: shall be the signature of the function being called. [...]
+    * ‘fnptrval’: An LLVM value containing a pointer to a function to be called. [...]
+
+One slightly surprising aspect of this instruction (at least, this was not
+obvious to me when I first read it) is that `fnptrval` does _not_ have the type
+`fnty`. Rather, `fnptrval` has a pointer type, and the type of the underlying
+memory has type `fnty`.
```
langston-barrett, line 1480: > the type of the underlying memory has type `fnty`

nit: I'm not sure how I feel about this statement. Can a chunk of executable code residing in memory be said to have some kind of function type, in the same sense that e.g, a struct stored on the heap has a struct type? Not exactly sure how to improve it either, so feel free to ignore this non-constructive comment :smile:

langston-barrett (approved):

src/Data/LLVM/BitCode/IR/Function.hs
RyanGlScott: I agree that this is a somewhat flimsy way to describe this. I was deliberately trying to avoid saying "the pointee type is `fnty`", as the phrase "pointee type" is not applicable for opaque pointers. On the other hand, I need a way to describe the property that `fnty` has. There must be _some_ way of doing this... I will see how LLVM's opaque pointers page talks about this.

RyanGlScott: The [opaque pointers page](https://llvm.org/docs/OpaquePointers.html) uses phrasing like this:

> Some instructions still need to know what type to treat the memory pointed to by the pointer as. For example, a load needs to know how many bytes to load from memory and what type to treat the resulting value as. In these cases, instructions themselves contain a type argument.

So LLVM's own phrasing isn't too far off from what I'm using above—they say the type to _treat_ the memory as, rather than the type _of_ the memory. Would you be OK with this framing, @langston-barrett?

RyanGlScott:
I have submitted elliottt/llvm-pretty#104 with the necessary changes to support pretty-printing `call` instructions properly in an opaque pointer setting. I will update the `llvm-pretty` submodule after that lands.

Because bumping the `llvm-pretty` submodule will require bringing in the changes from elliottt/llvm-pretty#103, I will most likely need to base this PR on top of #202 to make everything compile.

Issue #206, opened 2023-03-16
`disasm-test`: Run some tests only on certain LLVM versions


RyanGlScott:
Later LLVM versions often add new features to the bitcode language that cannot be parsed by earlier LLVM releases. We would still like to test these new features without having old versions of LLVM break when running `disasm-test`. To make this possible, we should institute a convention where files with certain suffixes—e.g., `mytest.llvm13.ll`—are only run with LLVM 13 or later, and skipped otherwise. This mirrors a similar convention used in `crux-llvm` (see [this code](https://github.com/GaloisInc/crucible/blob/1a7f06e192758c33d614edc8cf4f6fffdc8cea93/crux-llvm/test/Test.hs#L309-L312)), and we could reuse some of the code that it uses to check the versions of tools that you are using.

kquick:
Agreed, and this type of suffix-driven selectability is precisely the type of thing `tasty-sugar` is designed to help with.

RyanGlScott:
How does `tasty-sugar` work for a use case like `disasm-test`'s where there are no golden files? My understanding of how `disasm-test` is that for each `.ll` file, it does the following:

1. Run `llvm-as` on it to produce a bitcode file
2. Parse it back using `llvm-pretty-bc-parser`
3. Pretty-print it using `llvm-pretty`'s pretty-printer
4. Compare the pretty-printed output to the original `.ll` file using `diff` (after normalizing the output)

This process does not involve any golden files, as each `.ll` is compared against itself. I'm unclear on how to make `tasty-sugar` work in this situation. My first inclination was to try something like this:

```hs
cube :: TS.CUBE
cube = TS.mkCUBE { TS.inputDirs = ["disasm-test/tests"]
                 , TS.rootName = "*.ll"
                 , TS.separators = "."
                 , TS.expectedSuffix = "ll"
                 }
```

But this fails to find anything when passed to `findSugar`. I can only assume that this is because `expectedSuffix` clashes with `rootName`—if I change `expectedSuffix` to something else like `.ll-dummy` and copy each `.ll` file with a `.ll-dummy` suffix, then `findSugar` is able to find the `.ll` files. Of course, I really don't want to have to do this...

kquick:
Just released version 2.1.0.0 of tasty-sugar which should support the expected file being the same as the root file.

RyanGlScott:
[...]

Issue #167, opened 2021-09-13
Support parsing `poison` constant values


RyanGlScott:
LLVM 12 introduces a [`poison` constant value](https://releases.llvm.org/12.0.0/docs/LangRef.html#poison-values), similar in spirit to the `undef` constant value. Here is how to produce one:

```c
double f(int x) {
  return x % 0;
}
```
```
$ ~/Software/clang+llvm-12.0.0-x86_64-linux-gnu-ubuntu-20.04/bin/clang test.c -emit-llvm -O1 -fno-discard-value-names -S
test.c:2:12: warning: remainder by zero is undefined [-Wdivision-by-zero]
  return x % 0;
           ^ ~
1 warning generated.

$ ~/Software/clang+llvm-12.0.0-x86_64-linux-gnu-ubuntu-20.04/bin/clang test.c -emit-llvm -O1 -fno-discard-value-names -c
test.c:2:12: warning: remainder by zero is undefined [-Wdivision-by-zero]
  return x % 0;
           ^ ~
1 warning generated.

$ cat test.ll
; ModuleID = 'test.c'
source_filename = "test.c"
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
target triple = "x86_64-unknown-linux-gnu"

; Function Attrs: norecurse nounwind readnone uwtable willreturn
define dso_local double @f(i32 %x) local_unnamed_addr #0 {
entry:
  ret double poison
}

attributes #0 = { norecurse nounwind readnone uwtable willreturn "disable-tail-calls"="false" "frame-pointer"="none" "less-precise-fpmad"="false" "min-legal-vector-width"="0" "no-infs-fp-math"="false" "no-jump-tables"="false" "no-nans-fp-math"="false" "no-signed-zeros-fp-math"="false" "no-trapping-math"="true" "stack-protector-buffer-size"="8" "target-cpu"="x86-64" "target-features"="+cx8,+fxsr,+mmx,+sse,+sse2,+x87" "tune-cpu"="generic" "unsafe-fp-math"="false" "use-soft-float"="false" }

!llvm.module.flags = !{!0}
!llvm.ident = !{!1}

!0 = !{i32 1, !"wchar_size", i32 4}
!1 = !{!"clang version 12.0.0 (https://github.com/llvm/llvm-project/ b978a93635b584db380274d7c8963c73989944a1)"}
```

[...]

PR #210, opened 2023-03-23
`disasm-test`: Implement LLVM version–specific tests

RyanGlScott:
This copies over much of the version querying machinery from `crux-llvm`'s test suite so that `tasty-sugar` can pick tests depending on if you have a sufficiently recent version of `llvm-{as,dis}`. To test that it works as expected, I have added a belated test case for https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/167 that requires LLVM 12 or later (see `poison.at-least-llvm12.ll`).

Fixes https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/206.


kquick (approved):
Nice!

PR #220, opened 2023-04-24
`disasm-test`: Overhaul treatment of LLVM version ranges

RyanGlScott:
This is a minor rewrite of the logic used to check for LLVM version ranges in `disasm-llvm` test cases. The new approach is similar to Case 3/3a from this `tasty-sugar` document:
https://github.com/kquick/tasty-sugar/blob/1fc06bee124e02f49f6478bc1e1df13704cc4916/Ranges.org#case-3---explicit-and-a-weaker-match

In particular:

* We have adopted the convention that the test output for the most recent version of LLVM is always contained in a bare `.ll` file.
* There are no longer any `at-least-llvm*` files, just `pre-llvm*`. This greatly simplifies the number of cases to consider and the number of checks to implement.
* We now skip test configurations by having `SKIP_TEST` as the first line of the `.ll` file. Again, this greatly simplifies the logic needed to skip test cases on certain configurations.

This is heavily inspired by a similar change made in https://github.com/GaloisInc/crucible/pull/1083.


kquick (approved):

RyanGlScott:
The CI failure is due to https://github.com/fosskers/versions/issues/65. I'll restrict the upper version bounds on `versions` accordingly.

llvm-pretty-bc-parser.cabal
```diff
@@ -135,7 +135,7 @@ Test-suite disasm-test
                        tasty-hunit,
                        tasty-sugar >= 2.2 && < 2.3,
                        text,
-                       versions,
+                       versions < 6,
```
fosskers, line 138: Feel free to undo this, as I've just released `6.0.1`.

RyanGlScott: Thanks, @fosskers!

PR #221, opened 2023-05-03
More complete support for opaque pointers (required for LLVM 15+)

### LINKED ISSUE #177: Support opaque pointers

RyanGlScott:
This is a collection of various tweaks to `llvm-pretty-bc-parser` that allow it to fully support opaque pointers, an alternative representation of pointer types that is enabled by default in LLVM 15 or later. This patch should fix #177 insofar as I have tested these changes on a reasonable subset of LLVM 15/16 bitcode, and this is enough to make everything that I have thrown at `llvm-pretty-bc-parser` work.

I have split this patch up into self-contained commits, and for reviewing purposes, I'd recommend viewing each patch in isolation. The general themes of these patches are:

## Adapting to `llvm-pretty` API changes (from https://github.com/elliottt/llvm-pretty/pull/110)

* `Adapt to Load gaining an explicit load type`
* `Adapt to GEP/ConstGEP gaining explicit base types`

## Parsing opaque pointers

* `Basic support for parsing opaque pointer types`

## Making other instructions work with opaque pointers

* `Remove ptrTo and baseType`
* `Only use elimPtrTo on old LLVMs for invoke/call/callbr`
* `parseAtomicRMW: Don't inspect pointee type of argument on recent LLVMs`
* `cmpxchg: Compare argument types using eqTypeModuloOpaquePtrs`

## CI/documentation updates

* `README: Indicate support for LLVM 15 and 16`
* `llvm-quick-fuzz: Add LLVM 15 and 16 configurations`

## Test suite updates

* `disasm-test: Avoid llvm-as from crashing on mixed opaque/non-opaque pointers`
* `disasm-test: Rename callbr.ll to callbr.pre-llvm15.ll`
* `disasm-test: Add callbr test output for LLVM 15+`
* `disasm-test: Add opaque pointer test cases`


src/Data/LLVM/BitCode/Assert.hs
```diff
@@ -80,43 +79,37 @@ recordSizeIn record ns =
...
-    ty ->
-      fail $ unlines $
-        [ "Instruction expected a pointer argument."
-        , "Instruction signature: " ++ sig
-        , "Argument type: " ++ show ty
-        ]
+{-
+Note [Pointers and pointee types]
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Unlike LLVM itself, llvm-pretty and llvm-prety-bc-parser allow mixing opaque
```
kquick, line 105: misspelling: "prety"

src/Data/LLVM/BitCode/Assert.hs
```diff
@@ -80,43 +79,37 @@ recordSizeIn record ns =
...
+and non-opaque pointers. A consequence of this is that we generally avoid
+pattern matching on PtrTo (non-opaque pointer) types and inspecting the
+underlying pointee types. This sort of code simply won't work for PtrOpaque
+types, which lack pointee types.
+
+The elimPtrTo and elimPtrTo_ functions go against this rule, as they retrieve
+the pointee type in a PtrTo. These functions are primarily used for supporting
+old versions of LLVM which do not store the necessary type information in the
+instruction itself.
```
kquick, line 114: Is there a recommendation for what to use instead for opaque pointers and more modern approaches?

src/Data/LLVM/BitCode/Assert.hs
```diff
@@ -22,7 +22,6 @@ module Data.LLVM.BitCode.Assert
   -- ** Types
   , elimPtrTo
   , elimPtrTo_
-  , ptrTo
```
kquick, line 25: Can we provide some porting instructions (in this PR is fine) for code which might be using `ptrTo`?  What should be done in that code?

src/Data/LLVM/BitCode/IR/Function.hs
```diff
@@ -1133,20 +1111,11 @@ parseGEP t mbInBound r d = do
...
           (tv,ix') <- getValueTypePair t r' 2
-          -- TODO: the following sometimes fails, but it doesn't seem to matter.
-          {-
-          unless (baseType (typedType tv) == ty)
-              (fail $ unlines [ "Explicit gep type does not match base type of pointer operand"
-                              , "Declared type: " ++ show (ppType ty)
-                              , "Operand type: " ++ show (ppType (typedType tv))
-                              , "Base type of operand: " ++ show (ppType (baseType (typedType tv)))
-                              ])
-           -}
```
kquick, line 1144: Yikes!  Nice to get rid of this, albeit perhaps at the cost of less information available.

src/Data/LLVM/BitCode/Assert.hs
```diff
@@ -80,43 +79,37 @@ recordSizeIn record ns =
 ----------------------------------------------------------------
 -- ** Types
 
--- | Assert that this thing is a pointer, get the underlying type
+-- | Assert that this thing is a @'PtrTo' ty@ and return the underlying @ty@.
+--
+-- Think carefully before using this function, as it will not work as you would
+-- expect when the type is an opaque pointer.
+-- See @Note [Pointers and pointee types]@.
```
kquick, line 86: Should we add a WARNING pragma here and below (category "llvm-pointers" so that it can be explicitly disabled but is enabled by default)?  

See: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/pragmas.html#warning-and-deprecated-pragmas

kquick (approved):
Some documentation related comments, but overall looks good.

src/Data/LLVM/BitCode/Assert.hs
RyanGlScott: `elimPtrTo` is typically used in situations in which an explicit `Type` is needed, but the only way to retrieve the `Type` is by inspecting a pointer type. Because opaque pointers don't have pointee types, there is no direct `elimPtrTo` counterpart for opaque pointers. The modern approach would be to store the relevant `Type` information nearby so that it can always be retrieved independently of the pointer. There isn't an exact template that you can follow for this, but in the case of LLVM instructions, this is usually tantamount to having an extra `Type` field alongside the `Value` representing the pointer.

src/Data/LLVM/BitCode/Assert.hs
RyanGlScott: Yes, I can include some general advice about how to port code to adapt to opaque pointers. Come to think of it, we will need to make a new Hackage release of `llvm-pretty-bc-parser` at some point (see https://github.com/GaloisInc/crucible/issues/1077), so we should create a changelog and mention these changes there.

src/Data/LLVM/BitCode/Assert.hs
RyanGlScott: > Should we add a WARNING pragma here and below

I'm of two minds on this. While `elimPtrTo` is almost assuredly not what you want in an opaque pointer world, there are still some places where we need to use it in `llvm-pretty-bc-parser` for backwards compatibility with old LLVMs. As a result, we would have to disable deprecation warnings in `llvm-pretty-bc-parser` wholesale to avoid this. 

> category "llvm-pointers" so that it can be explicitly disabled but is enabled by default

The category feature is cool, and this would give us a way to avoid the disabling-warnings-wholesale problem mentioned above. Sadly, warning categories are a feature that is only available in GHC 9.8 (not yet released) or later. We could add a category, but it wouldn't be useful to us for some time.

kquick:
Dang, I missed that the category feature for user warnings was a new item.

I did a scan of most of our code, and the only calls I can find to `elimPtrTo` are only in llvm-pretty and llvm-pretty-bc-parser, so  I'm much less concerned about not having warnings (or even migration documentation) at this point.

Issue #222, opened 2023-06-20
Debug information incorrect


kquick:
With the change from https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/202/commits/68d26a5f22f1fe8226fdb5863793ab0df7879a13 there is a regression that affects debug information.  This was noticed in a client of this library when file and line information for functions was no longer available and tracked down to the `guard` statement in this identified commit.

It appears that this causes some internal structure synchronization issues of some kind.  I will work on the test case more, but I added a smallprog.c file (compile via `clang` with `-c -emit-llvm -O0 -g` and then use `llvm-disasm`) which exhibits this metadata issue.  In the previous implementation (without the guard), the following metadata is observed:

```
...
!28 = !{!"clang version 12.0.1"}
!29 = !{!"llvm.loop.mustprogress"}
!30 = !DIDerivedType(tag: 15, line: 0, baseType: !9, size: 64, align: 0, offset: 0, flags: 0)
!31 = !{null, !30}
!32 = distinct !DISubroutineType(flags: 0, types: !31)
!33 = distinct !DISubprogram(scope: !4, name: "sp_receive", file: !4, line: 17, type: !41, isLocal: false, isDefinition: false, scopeLine: 17, virtuality: 0, virtualIndex: 0, flags: 0, isOptimized: false, unit: !3, retainedNodes: !5)
!34 = distinct !DILexicalBlock(scope: !33, file: !4, line: 18, column: 24)
!35 = distinct !DILexicalBlock(scope: !34, file: !4, line: 20, column: 9)
!36 = distinct !DILexicalBlock(scope: !35, file: !4, line: 20, column: 9)
!37 = distinct !DILexicalBlock(scope: !36, file: !4, line: 20, column: 53)
...
```

With the newer code with the `guard` statement in place, the output changes:

```
...
!28 = !{!"clang version 12.0.1"}
!29 = !{!"llvm.loop.mustprogress"}
!30 = !DIDerivedType(tag: 15, line: 0, baseType: !9, size: 64, align: 0, offset: 0, flags: 0)
!31 = !{null, !30}
!32 = distinct !DISubprogram(scope: !4, name: "sp_receive", file: !4, line: 17, type: !40, isLocal: false, isDefinition: false, scopeLine: 17, virtuality: 0, virtualIndex: 0, flags: 0, isOptimized: false, unit: !3, retainedNodes: !5)
[...]

PR #223, opened 2023-06-20
Revert metadata change causing Issue #222

kquick:
This reverts the recent metadata change (related to LLVM-13 changes) that caused Issue #222 until the underlying problems caused by that change can be identified and handled properly.

There is a new file (diasm-test/cpp/smallprog.c) that demonstrates the issue.  This should be integrated into self-tests, and there should be corresponding self-tests that affirm the needed functionality that this reverts when that is re-established.

### DIFF
--- a/disasm-test/cpp/smallprog.c
+++ b/disasm-test/cpp/smallprog.c
@@ -1,0 +1,42 @@
+/* #include <stdint.h> */
+
+#define MESSAGE_CAPACITY 5
+
+typedef struct message {
+    int type;
+    int* payload;
+    int payload_len;
+} message_t;
+
+typedef struct broker {
+    message_t msgs[MESSAGE_CAPACITY];
+    int msg_idx;
+    int num_msgs;
+} broker_t;
+
+void sp_receive(message_t* msg) {
+    switch (msg->type) {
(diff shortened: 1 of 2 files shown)


disasm-test/cpp/smallprog.c
```diff
@@ -0,0 +1,42 @@
+/* #include <stdint.h> */
```
RyanGlScott, line 1: Silly question: what actually tests this code? I was under the impression that `disasm-test` was ingesting `.ll` files, not `.c` files, so the lack of a corresponding `smallprog.ll` file surprises me.

kquick: I hadn't generate an `.ll` file: I assumed that it might be different for different clang versions.

Currently this isn't part of any tests; I probably should have marked this PR as draft, my main goal was to drive awareness of this regression but I haven't had time to dive into it further.  I need to take care of a couple of other tasks first, and hopefully I will be able to extend this soon, unless anyone else happens to get there first.

RyanGlScott:
I propose an alternate approach in #228.

RyanGlScott:
Subsumed by #228.

PR #225, opened 2023-06-26
Support building with GHC 9.6

### LINKED ISSUE #224: Build with GHC 9.4, 9.6
There are two more recent major versions of GHC available: https://www.haskell.org/ghc/download.html

RyanGlScott:
This makes two changes to `llvm-pretty-bc-parser` to make it build with GHC 9.6:

* Because `mtl-2.3.*` no longer re-exports `Control.Monad` or `Control.Monad.Fix`, I needed to tighten up the imports in `Data.LLVM.BitCode.Parse` to make it build with GHC 9.6, which bundles `mtl-2.3.1`.
* Bump the `llvm-pretty` submodule to bring in the changes from https://github.com/elliottt/llvm-pretty/pull/112.

This also updates the `README` and CI to test against GHC 9.2, 9.4, and 9.6 (the three most recent GHC releases at the time of writing).

Fixes #224.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,7 @@ jobs:
       matrix:
         os: [ubuntu-22.04]
         # See doc/developing.md
-        ghc: ["8.10.7", "9.0.2", "9.2.4"]
+        ghc: ["9.2.8", "9.4.5", "9.6.2"]
     name: build - ${{ matrix.ghc }} - ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/llvm-quick-fuzz.yml
+++ b/.github/workflows/llvm-quick-fuzz.yml
@@ -22,7 +22,7 @@ jobs:
       matrix:
         os: [ubuntu-22.04]
         # See doc/developing.md
-        ghc: ["8.10.7"]
+        ghc: ["9.2.8"]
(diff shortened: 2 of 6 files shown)


RyanGlScott:
Eek. The fuzzing harness fails to build on GHC 9.6 because it depends on `monad-par`, and [`monad-par` doesn't currently build with GHC 9.6](https://github.com/simonmar/monad-par/pull/75)...

langston-barrett (commented):
LGTM once the dependency can be upgraded, may want to update here as well: https://github.com/GaloisInc/llvm-pretty-bc-parser/blob/master/llvm-pretty-bc-parser.cabal#L11C5-L11C5

RyanGlScott:
Relatedly, I wonder if this `monad-par` dependency is even worth the bother. In principle, it's used for parallelizing an invocation of the fuzzer across multiple `clang` versions, but the default setting just uses a single `clang`. Not even the CI makes use of this option, and if CI isn't using it, then I'd be surprised if anyone is using it.

kquick (approved):
I'm in agreement with the de-monad-paritization of the fuzz tool and moving forward on this PR.

RyanGlScott:
Successfull `llvm-quick-fuzz` workflow with these changes: https://github.com/GaloisInc/llvm-pretty-bc-parser/actions/runs/5390582172

PR #226, opened 2023-06-27
Llvm regressions

kquick:
This establishes another channel of CI using nix, which will be used in the future to run tests against multiple versions of llvm and clang.  This also updates some of the test code for easier and faster compilation.

### DIFF
--- a/.github/workflows/nix-ci.yml
+++ b/.github/workflows/nix-ci.yml
@@ -28,6 +28,7 @@ jobs:
 
   tests:
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     steps:
       - uses: cachix/install-nix-action@v20
         with:
--- a/.github/workflows/nix-ci.yml
+++ b/.github/workflows/nix-ci.yml
@@ -9,6 +9,9 @@ on:
 
 jobs:
   build:
+    # This job builds the main package, via a number of GHC versions.  This is
+    # the build that would be used by downstream packages using this as a
+    # dependency; this does not build documentation or tests.
     runs-on: ubuntu-latest
     strategy:
       matrix:
@@ -27,6 +30,7 @@ jobs:
         run: nix build -L github:${{ github.repository }}/${{ github.sha }}#llvm-pretty-bc-parser.${{ matrix.ghc-version }}
 
   tests:
+    # This job builds the tests, and then runs the tests (in two different steps)
     runs-on: ubuntu-latest
     steps:
       - uses: cachix/install-nix-action@v20
@@ -45,6 +49,7 @@ jobs:
         run: nix build -L github:${{ github.repository }}/${{ github.sha }}#TESTS
 
   doc:
+    # This job builds the documentation (e.g. Hackage, etc.)
     runs-on: ubuntu-latest
     steps:
       - uses: cachix/install-nix-action@v20
--- a/.github/workflows/nix-ci.yml
+++ b/.github/workflows/nix-ci.yml
@@ -28,7 +28,6 @@ jobs:
 
   tests:
     runs-on: ubuntu-latest
-    timeout-minutes: 30
     steps:
       - uses: cachix/install-nix-action@v20
         with:
--- a/.github/workflows/nix-ci.yml
+++ b/.github/workflows/nix-ci.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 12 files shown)


.github/workflows/nix-ci.yml
```diff
@@ -0,0 +1,61 @@
...
+          name: galois
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+      - name: test build
+        shell: bash
+        run: nix build github:${{ github.repository }}/${{ github.sha }}#TESTS_PREP
+      - name: test
+        shell: bash
+        run: nix build -L github:${{ github.repository }}/${{ github.sha }}#TESTS
+
+  doc:
```
RyanGlScott, line 48: What does this job do?

.github/workflows/nix-ci.yml
```diff
@@ -0,0 +1,61 @@
...
+      - 'release-**'
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        ghc-version: [ "ghc94", "ghc92", "ghc810", "ghc88" ]
```
RyanGlScott, line 15: It's interesting that this has an entirely separate build matrix from `build.yml`. My understanding is that this is primarily meant to test different LLVM versions against `disasm-test`, so is there much of a point to trying this step with different GHC versions?

flake.nix
```diff
@@ -0,0 +1,156 @@
...
+              buildInputs = [ llvm pkgs.diffutils pkgs.coreutils ];
+            };
+        in rec {
+          default = llvm-pretty-bc-parser;
+          TESTS = wrap "llvm-pretty-bc-parser-TESTS"
+            (builtins.map
+              (llvm-pretty-bc-parser-test llvm-pretty-bc-parser-test-build)
+              [
+                pkgs.llvm_10
+                pkgs.llvm_11
```
RyanGlScott, line 107: This is where all of the different LLVM versions are enumerated, right? It might be worth stating this somewhere in a comment—I had to trawl through the CI logs before I realized which LLVM versions were actually being tested.

Relatedly, is there any reason we couldn't go ahead and add more versions here?

RyanGlScott (commented):
I like the direction this is going in. I have some questions about specifics.

.github/workflows/nix-ci.yml
kquick: I added some comments to the CI yaml file.  Basically, the primary `build` job only builds the code (only) for different GHC versions, the `tests` job first builds the tests (for a default GHC version) and then runs the tests.  The `doc` job builds any documentation (e.g. Haddock).

.github/workflows/nix-ci.yml
kquick: It wasn't hard to add this and seemed like a good backup to the other build channel, even though it's somewhat of a duplication.  This is also a replication of a similar CI yaml file I've used elsewhere where those locations don't have a parallel path, so this keeps the diff cleaner between those and this.

RyanGlScott: > even though it's somewhat of a duplication

Indeed, `nix-ci.yml` does nearly everything that `build.yml` does, plus additional things. I'm wondering if we shouldn't just remove `build.yml` in favor of `nix-ci.yml`.

flake.nix
kquick: Hrm.. I thought I had replied to this, but I don't see it now: apologies if this ends up being a duplicate.

Short version is: there are issues with both lower and higher LLVM versions that will probably want accompanying fixes (in both code and modifications to testing).  These seem to be the two "stable"/"baseline" versions that are good to start with, and this PR was already getting large, so I decided to leave this range expansion for future PRs.

kquick: And yes, this is the location where the LLVM versions are specified and I added a commit with descriptive comments.

.github/workflows/nix-ci.yml
kquick: It's a possibility.  I'd suggest deferring that decision a little until I get some of these regression tests added and then make this determination.  I am also probably not the best qualified to vote on this, being very comfortable with the nix approach which is not necessarily the case for others, and the `build.yml` is more conventional.

RyanGlScott: Fair enough.

RyanGlScott (approved):

PR #230, opened 2023-07-03
Add c disasm test

kquick:
The disasm-test will now check .c files by compiling with the available version of clang and ensuring that the result can be assembled and disassembled properly.  This provides additional verification over the use of existing .ll sources because the output from clang will contain changes associated with changes in LLVM versions whereas the .ll sources are more static.

Also adds a test to ensure that the metadata generated has unique indices.

Adds a source to demonstrate the failure of the above.

Uses tasty-expected-failure to show both llvm-version skipped tests and to disable (for now) test failures for newly failing tests.


disasm-test/Main.hs
```diff
@@ -578,3 +735,14 @@ rmFile tmp = do Keep keep <- gets keepTemp
...
+
+knownBugs :: [ (FilePath, String) ]  -- FilePath is TS.rootMatchName
+knownBugs =
+  let pr223pr228 = "unnamed metadata indices are currently duplicated (see https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/223 and https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/228)"
+  in
+    [
+      ( "p0.ll",  pr223pr228 )
+    , ( "p0.c",   pr223pr228 )
+    , ( "T189.c", pr223pr228 )
+    , ( "callbr.c", "label targets not pretty-printed correctly, causing llvm-as failure" )
```
RyanGlScott, line 747: Is there an issue for this somewhere?

disasm-test/Main.hs
```diff
@@ -578,3 +735,14 @@ rmFile tmp = do Keep keep <- gets keepTemp
                   $ do Details dets <- gets showDetails
                        when dets $ liftIO $ putStrLn $ "## Removing " <> tmp
                        liftIO $ removeFile tmp
+
+knownBugs :: [ (FilePath, String) ]  -- FilePath is TS.rootMatchName
+knownBugs =
+  let pr223pr228 = "unnamed metadata indices are currently duplicated (see https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/223 and https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/228)"
```
RyanGlScott, line 741: Consider breaking this up into multiple lines for easier readability.

disasm-test/Main.hs
```diff
@@ -512,6 +652,23 @@ processBitCode pfx file = do
...
+postParseTests m = ensureValidMetadataIndices m
+  where
+    ensureValidMetadataIndices md = do
+      let idxs = AST.umIndex <$> AST.modUnnamedMd md
+      let uniqIdxs = nub idxs
+      let numDups = length idxs - length uniqIdxs
+      unless (numDups == 0)
+        $ do Details det <- gets showDetails
+             when det $ liftIO
+               $ putStrLn $ "Unnamed MetaData indices: " <> show idxs
```
RyanGlScott, line 666: I'm not sure if I've ever seen "metadata" capitalized in CamelCase before. I'd propose just using:

```suggestion
               $ putStrLn $ "Unnamed metadata indices: " <> show idxs
```

disasm-test/Main.hs
```diff
@@ -512,6 +652,23 @@ processBitCode pfx file = do
...
+             liftIO $ assertFailure
+               $ show numDups <> " duplicated Unnamed MetaData indices"
```
RyanGlScott, line 668:
```suggestion
               $ show numDups <> " duplicated Unnamed metadata indices"
```

disasm-test/Main.hs
```diff
@@ -512,6 +652,23 @@ processBitCode pfx file = do
         when det $ liftIO $ putStrLn $ "## parsed Bitcode to " <> parsed
         return (parsed, Nothing)
 
+
+postParseTests :: AST.Module -> TestM ()
```
RyanGlScott, line 656: This could use a comment explaining what it is checking for.

disasm-test/Main.hs
```diff
@@ -318,119 +361,200 @@ cube = TS.mkCUBE
...
+compilerCube :: TS.CUBE
+compilerCube = assemblyCube
+               { TS.rootName = "*.c"
+               , TS.sweetAdjuster = rangeMatch
+               }
+
+
+runCompileTest :: TS.Sweets -> TS.Expectation -> IO [TestTree]
+runCompileTest sweet expct = do
+  skipTest <- ("SKIP_TEST" `L.isPrefixOf`) <$> L.readFile (TS.expectedFile expct)
```
RyanGlScott, line 493: I'm slightly bothered by the fact that this magic string `"SKIP_TEST"` is now used in multiple locations. Can you factor out the `"SKIP_TEST"` logic into its own function?

disasm-test/Main.hs
```diff
@@ -318,119 +361,200 @@ cube = TS.mkCUBE
...
+                 diffCmp ast1 ast2 -- Ensure that the ASTs match
+
+                 -- Ensure that the disassembled files match.  This is usually
+                 -- too strict (and doesn't really provide more info).  We
+                 -- normalize the AST (see below) to ensure that the ASTs match
+                 -- modulo metadata numbering, but the equivalent isn't possible
+                 -- for the assembly: we need llvm-as to be able to re-assemble
+                 -- it.
+                 --
+                 -- diffCmp parsed1 parsed2
```
RyanGlScott, line 420: Is this a stray comment? There is another use of `diffCmp` directly above this, and I'm not sure if the comment here is meant to apply to that line or not.

disasm-test/Main.hs
```diff
@@ -318,119 +361,200 @@ cube = TS.mkCUBE
...
+-- Compiler->Assembly->Disassembly tests
+
+-- The compilation tests ensure that the clang version-specific generated .bc
+-- file can be reasonably parsed by this library.  This is a parallel to the
+-- assemblyCube-driven tests, but starts with a C source file.  One distinction
+-- is that the .ll used for the assemblyCube is typically representative of a
+-- specific LLVM version, and while it is assembled and disassembled by newer
+-- versions of LLVM tools, it will never introduce any newer element, whereas the
+-- clang-generated bitcode will contain version-current output which might have
+-- newer elements and ordering.
```
RyanGlScott, line 477: How fragile is this? I worry that new Clang versions might introduce minor differences in LLVM instructions that, while not changing the user-observable behavior of the program, would nevertheless cause this `diff` test to fail. Do you expect that we will have to add a lot of additional, version-specific `.ll` expected files over time?

disasm-test/Main.hs
kquick: It is intentional: it's talking about comparing  the .ll files (represented by `parsed1` and `parsed2`), as opposed to the serialized AST (via `ppLLVM`) above that.

disasm-test/Main.hs
kquick: Done.

disasm-test/Main.hs
kquick: I was trying to match the internal variable name, but missed.  I've clarified these references.

disasm-test/Main.hs
kquick: Added.

disasm-test/Main.hs
kquick: Good idea: done.

disasm-test/Main.hs
kquick: There are definitely issues discovered using this method that are not seen with the `.ll` sources.  This may require adding new version-specific .ll target expected files in the future. The primary goal here is to be able to parse any bitcode file generated by clang (or any other llvm front-end); we are currently measuring  the success of that by determining if we can emit the parsed form in a format that the `llvm-as` tool will successfully parse back into bitcode, and then repeating that cycle to ensure that the two forms match.  I agree we may have to deal with format change issues, but at present I don't see a better way to test this, and adding these clang tests has revealed at least 5 bugs (not all are present in this PR: I wanted to establish the testing framework and then try to file them individually after this is in place; the two noted here are already present from existing `.c` files in the test corpus).

RyanGlScott: OK. My feeling on the matter is that we should keep the size of these `*.c` programs as small as possible to reduce the chance of future Clang versions compiling things in an ever-so-slightly different way and breaking the diff. It might be worth making a note of this general guideline in the [`llvm-disasm` `README`](https://github.com/GaloisInc/llvm-pretty-bc-parser/blob/df5b7d63cd6d7b06ecd13d9ce8eb189add2ab90a/disasm-test/README.md).

disasm-test/Main.hs
kquick: Added an issue.

disasm-test/Main.hs
kquick: I'm in agreement, and I updated the descriptive text in the disasm-text (there didn't seem to be a great spot in the README).

disasm-test/Main.hs
```diff
@@ -578,3 +749,19 @@ rmFile tmp = do Keep keep <- gets keepTemp
...
                        liftIO $ removeFile tmp
+
+knownBugs :: [ (FilePath, String) ]  -- FilePath is TS.rootMatchName
+knownBugs =
+  let pr223pr228 =
+        "unnamed metadata indices are currently duplicated \
+        \ (see https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/223 \
+        \ and https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/228)"
+      issue231 =
+        "label targets not pretty-printed correctly, causing llvm-as failure"
```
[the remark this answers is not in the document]
kquick: Done in #232

PR #234, opened 2023-07-07
Ensure test failures cause CI failures; add C++ test sources

kquick:
* Fixes the nix flake-based testing to cause the test run to fail when any of the individual tests failed.
* Adds the ability to have .cc input files (C++) for disasm-test runs, and ensures clang/clang++ is available when running the tests.
* Informational output and diffs generated bi disasm-test now ignore blank lines and whitespace in diff reporting.  This has no effect on whether the test passes or fails, it is simply diagnostic information reported to the user.

### DIFF
--- a/disasm-test/Main.hs
+++ b/disasm-test/Main.hs
@@ -479,7 +479,15 @@ parseBC pfx bc = do
     Details dets <- gets showDetails
     when dets $ liftIO $ do
       -- Informationally display if there are differences between the llvm-dis
-      -- and llvm-disasm outputs, but no error if they differ.
+      -- and llvm-disasm outputs, but no error if they differ.  Note that the
+      -- arguments to this diff are not the same as those supplied to the diffCmp
+      -- function. The diff here is intended to supply additional information to
+      -- the user for diagnostics, and whitespace changes are likely unimportant
+      -- in that context; this diff does not determine the pass/fail status of
+      -- the testing.  On the other hand, the diffCmp *does* determine if the
+      -- tests pass or fail, so ignoring whitespace in that determination would
+      -- potentially weaken the testing to an unsatisfactory degree and would
+      -- need more careful evaluation.
       putStrLn "## Output differences: LLVM's llvm-dis <--> this llvm-disasm"
       ignore (Proc.callProcess "diff" ["-u", "-b", "-B", "-w", norm, parsed])
       putStrLn ("successfully parsed " ++ show pfx ++ " bitcode")
--- a/disasm-test/Main.hs
+++ b/disasm-test/Main.hs
@@ -476,7 +476,8 @@ parseBC pfx bc = do
     when dets $ liftIO $ do
       -- Informationally display if there are differences between the llvm-dis
       -- and llvm-disasm outputs, but no error if they differ.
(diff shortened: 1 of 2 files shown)


disasm-test/Main.hs
```diff
@@ -592,7 +602,8 @@ assembleToBitCode pfx file = do
 compileToBitCode :: FilePath -> FilePath -> TestM FilePath
 compileToBitCode pfx file = do
   tmp <- liftIO getTemporaryDirectory
-  Clang comp <- gets clang
+  Clang comp' <- gets clang
+  let comp = if ".cc" `isSuffixOf` file then comp' <> "++" else comp'
```
RyanGlScott, line 606: There are some other language extensions that we could consider checking for, such as `.cpp`, `.cxx`, and `.C`. (List taken from [here](https://github.com/GaloisInc/crucible/blob/2cf85d801d806a5f62cfc19403e041169be66cb7/crux-llvm/src/Crux/LLVM/Compile.hs#L48-L50).)

disasm-test/Main.hs
```diff
@@ -492,16 +497,21 @@ parseBC pfx bc = do
...
-               , TS.sweetAdjuster = rangeMatch
-               }
+cCompilerCube :: TS.CUBE
+cCompilerCube = assemblyCube
+                { TS.rootName = "*.c"
+                , TS.sweetAdjuster = rangeMatch
+                }
+
+ccCompilerCube :: TS.CUBE
+ccCompilerCube = cCompilerCube { TS.rootName = "*.cc"}
```
RyanGlScott, line 514: Ah, but you're specifically checking against `*.cc` here... I'm not sure if `tasty-sugar` easily supports having multiple file extensions in the `rootName`. If not, feel free to ignore this suggestion.

disasm-test/Main.hs
```diff
@@ -476,7 +480,8 @@ parseBC pfx bc = do
     when dets $ liftIO $ do
       -- Informationally display if there are differences between the llvm-dis
       -- and llvm-disasm outputs, but no error if they differ.
-      ignore (Proc.callProcess "diff" ["-u", norm, parsed])
+      putStrLn "## Output differences: LLVM's llvm-dis <--> this llvm-disasm"
+      ignore (Proc.callProcess "diff" ["-u", "-b", "-B", "-w", norm, parsed])
```
RyanGlScott, line 484: This passes different flags to `diff` as compared to its other use in `diffCmp`. Should these flags be the same in both call sites?

kquick: This diff is only displayed to the user (when the new `--details` flag is specified) and isn't used for any actual success or failure determination.  The other diff *is* part of that determination; perhaps it should be tolerant of whitespace changes, but that would need to be pretty carefully evaluated, whereas my goal here was  to allow the user to focus on significant changes rather than the whitespace "noise".

RyanGlScott: I'll leave it to you to decide how this `diff` should behave, but I would recommend commenting why this `diff` ignores whitespace, since this is seemingly an intentional difference from how `diffCmp` works.

RyanGlScott (approved):

PR #236, opened 2023-08-04
Unify the conversion of Symbol to user-presented information.

kquick:
This method of converting a Symbol into a user-readable representation may change, so this consolidates it into a single location for easier future updates.

### DIFF
--- a/src/Data/LLVM/BitCode/IR/Function.hs
+++ b/src/Data/LLVM/BitCode/IR/Function.hs
@@ -31,6 +31,21 @@ import qualified Data.Map as Map
 import qualified Data.Sequence as Seq
 import qualified Data.Traversable as T
 
+
+-- When showing a Symbol to the user, show it in the manner that it would appear
+-- in LLVM text format.
+--
+-- NOTE: this cannot be eta-reduced to point-free format because implicit
+-- parameters are not supported in in point-free format.
+--
+-- TODO: this shows it in the manner that LLVM v3.5 would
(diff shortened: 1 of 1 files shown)


src/Data/LLVM/BitCode/IR/Function.hs
```diff
@@ -31,6 +31,21 @@ import qualified Data.Map as Map
 import qualified Data.Sequence as Seq
 import qualified Data.Traversable as T
 
+
+-- When showing a Symbol to the user, show it in the manner that it would appear
+-- in LLVM text format.
+--
+-- NOTE: this cannot be eta-reduced to point-free format because implicit
+-- parameters are not supported in in point-free format.
```
RyanGlScott, line 39: This isn't true. The real reason you can't write this point-free is because GHC 9.0+ implements [simplified subsumption](https://downloads.haskell.org/ghc/9.6.2/docs/users_guide/exts/rank_polymorphism.html#subsumption), which means that some applications of higher-rank functions (e.g., `ppLLVM`) have to be eta expanded in order to typecheck.

RyanGlScott (approved):
LGTM, aside from one quibble about documentation.

src/Data/LLVM/BitCode/IR/Function.hs
kquick: Thanks, I had been aware of the eta expansion needs but hadn't mapped it to this particular situation.  I've updated this.

PR #237, opened 2023-08-04
Use llvm version in testing

kquick:
This passes the llvm version down to the testing functions which allows them to generate the `.ll` text format that corresponds to the current llvm tools.


disasm-test/Main.hs
```diff
@@ -675,7 +680,24 @@ processBitCode pfx file = do
...
+            case v of
+              3 -> case vcVersioning llvmVersion ^? (_Right . minor) of
+                     Just 5 -> return $ ppLLVM35 $ llvmPP m'
+                     Just 6 -> return $ ppLLVM36 $ llvmPP m'
+                     Just 7 -> return $ ppLLVM37 $ llvmPP m'
+                     Just 8 -> return $ ppLLVM38 $ llvmPP m'
+                     o -> if maybe True (< 5) o
+                          then return $ ppLLVM35 $ llvmPP m'
+                          else return $ ppLLVM38 $ llvmPP m'
+              _ -> return $ ppLLVM38 $ llvmPP m'  -- TODO: LLVM38 is the last one currently defined
```
RyanGlScott, line 699: This version-matching logic ought to be in its own function, it for no other reason than it adds a significant amount of clutter to `processBitCode`.

RyanGlScott (approved):

disasm-test/Main.hs
kquick: I'm headed that direction, just trying to make sure I'm capturing all the use-case patterns before I consolidate this.

PR #244, opened 2023-08-09
Fix attachment of instruction metadata in blocks.

kquick:
Fix to correctly attach instruction metadata (the `!llvm.loop !3`) when parsing function blocks.

### DIFF
--- a/disasm-test/tests/instrmd.ll
+++ b/disasm-test/tests/instrmd.ll
@@ -1,0 +1,33 @@
+define i32 @f(i32 %a0) {
+   %1 = mul i32 %a0, 2
+   %2 = mul i32 %1, 2
+   %3 = mul i32 %2, 2
+   br label %test
+test:
+   %4 = mul i32 %a0, 2, !llvm.loop !3
+   br label %5, !dbg !1
+; <label>:5
+   %6 = mul i32 %4, 2
+   %7 = mul i32 %6, 2
+   %8 = mul i32 %7, 2
+   %9 = mul i32 %8, 2
+   %10 = mul i32 %9, 2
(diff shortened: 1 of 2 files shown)


RyanGlScott:
What issue does this fix? (I can believe that the old code was wrong, but it's not obvious to me at a glance what this is trying to address, nor why it is the correct fix.)

kquick:
The `!llvm.loop !3` portion of lines 7 and 17 in the added test demonstrate "instruction metadata".  This metadata is collected separately and then attached to the instructions.  In this particular case, since the `off` wasn't being included, it would never see the correct instruction offset (except for the first block of the function) and would therefore not attach the metadata.  The manifestation of this is that when parsing the bitcode file, these instruction metadata fields would be dropped, and subsequently no longer be attached to the instructions.  The added test file exercises this via the round-trip, where the comparison of the round-trip result lacks that additional metadata (visible if you run the added test without the fix in `Function.hs`).

RyanGlScott:
Thanks, that helps clarify the bug. (In the future, it would be helpful to open a GitHub issue about things like this, since it helps to have an issue describe the problem, following by a PR which implements a solution to the problem.)

How did you arrive at the conclusion that this was the proper fix? Was there any code in the upstream LLVM repo that you took inspiration from that we could cite here as justification for this code? If not, that's fine, but I do want to record something if possible for the sake of bringing future readers of this code up to speed more quickly.

RyanGlScott (approved):
Hm, OK. The offset code is still a little mysterious to me, but the new `i-off` code is consistent with the `applies (i,_) = i - off < numStmts` check elsewhere in the code. I'll go ahead and approve this.

PR #249, opened 2023-08-15
Add ability to test based on pre-generated bitcode files.

kquick:
Allows testing to be driven from raw bitcode (`.bc`) files to test the permissive capabilities of the parser (see PR #247 ).


disasm-test/Main.hs
```diff
@@ -558,6 +562,45 @@ runCompileTest llvmVersion knownBugs sweet expct = do
...
+            case ast of
+              Nothing ->
+                -- No round trip, so this just verifies that the bitcode could be
+                -- parsed without generating an error.
+                return ()
+              Just ast1 ->
+                -- Assemble and re-parse the bitcode to make sure it can be
+                -- round-tripped successfully.
+                with2Files (processLL pfx parsed1)
+                $ \(_, Just ast2) -> diffCmp ast1 ast2
```
RyanGlScott, line 600: This uni-pattern match on `Just` emits a warning with GHC 9.2 or later:

```
disasm-test/Main.hs:600:19: warning: [-Wincomplete-uni-patterns]
    Pattern match(es) are non-exhaustive
    In a lambda abstraction:
        Patterns of type ‘(FilePath, Maybe FilePath)’ not matched:
            ([], Nothing)
            ((_:_), Nothing)
    |
600 |                 $ \(_, Just ast2) -> diffCmp ast1 ast2
    |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Actually, two other uses of `with2Files` in this file produce the same warning as well. Is it possible to restructure things to avoid this?

RyanGlScott (approved):
Awesome, thanks! One question below, but otherwise this LGTM.

disasm-test/Main.hs
kquick: Sure: fixed.

Issue #250, opened 2023-09-23
parsing fails on cmpxchg instructions


kquick:
Bitcode files which used to parse are now failing with the following error:

```
Mismatched value types:
cmp value: ValIdent (Ident "52")
new value: ValIdent (Ident "53")
cmp type:  PrimType (Integer 32)
new type:  PrimType (Integer 32)

from:
        FUNC_CODE_INST_CMPXCHG
        @_ZN18LockstepComponents18register_componentEv
        FUNCTION_BLOCK
        FUNCTION_BLOCK_ID
        value symbol table
        MODULE_BLOCK
        Bitstream
```

PR #228, opened 2023-07-03
Don't put DIExpression/DIArgList into mtNodes

RyanGlScott:
Instead, put them in `mtEntries`, which aren't used when populating top-level metadata lists. For more information, see the new `Note [Printing metadata inline]`.

Fixes #222. Subsumes #223.

PR #251, opened 2023-09-23
Fix logic in updating cmpxchg parsing for opaque pointers.

kquick:
Fixes: https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/250
Caused by: https://github.com/GaloisInc/llvm-pretty-bc-parser/commit/2e848a63ae9661e32e919c8177399ff06bf2040a
Added test requires: https://github.com/elliottt/llvm-pretty/pull/129, https://github.com/elliottt/llvm-pretty/pull/127, and https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/228


RyanGlScott (approved):
An embarrassing oversight on my part. Thanks for fixing this!

PR #252, opened 2023-09-24
Update parsing of METADATA_SUBRANGE for new and changed DISubrange fields.

kquick:
Requires https://github.com/elliottt/llvm-pretty/pull/130


src/Data/LLVM/BitCode/IR/Metadata.hs
```diff
@@ -534,15 +534,23 @@ parseMetadataEntry vt mt pm (fromEntry -> Just r) =
...
-      -- files generated by newer Clang versions, but it won't crash.
-      diNode         <- DISubrange
-        <$> parseField r 1 numeric     -- disrCount
-        <*> parseField r 2 signedInt64 -- disrLowerBound
+      field0 <- parseField r 0 unsigned
+      let isDistinct = field0 .&. 0 == 1
+      let format = field0 `shiftR` 1
+      ctx <- getContext
+      diNode <- case format of
+        2 -> do count <- mdForwardRefOrNull ctx mt <$> parseField r 1 numeric
```
RyanGlScott, line 542: This `format` logic is tricky. It would be worth citing the relevant part of the LLVM codebase ([here](https://github.com/llvm/llvm-project/blob/d1c643a0168268445dab29d96b96d5fcb9c97f92/llvm/lib/Bitcode/Reader/MetadataLoader.cpp#L1437-L1445), I believe) from which this takes inspiration.

kquick: Yup: added.

kquick:
Updated to match the recent changes in https://github.com/elliottt/llvm-pretty/pull/130 where the `count` and `lowerBound` of `DISubrange` are always represented as a `Maybe ValMd` encoding instead of an `Either Int64 (Maybe ValMd)`

RyanGlScott (approved):
LGTM (once https://github.com/elliottt/llvm-pretty/pull/130 lands upstream and is incorporated here, of course).

PR #253, opened 2023-09-26
Fix parsing for DwarfAddressSpace for DIDerivedType.

kquick:
In the bitcode, the address space is encoded as +1, where 0 means there is no address space.  This parsing will store the actual address space in the AST and use `Nothing` to indicate the "no address space" condition.  Note that this latter is indistinguishable from "not specified" for LLVM 4 and earlier, but this is not expected to be a significant distinction.

Supercedes https://github.com/elliottt/llvm-pretty/pull/125.

### DIFF
--- a/src/Data/LLVM/BitCode/IR/Metadata.hs
+++ b/src/Data/LLVM/BitCode/IR/Metadata.hs
@@ -614,7 +614,8 @@ parseMetadataEntry vt mt pm (fromEntry -> Just r) =
              then pure Nothing  -- field not present
              else do v <- parseField r 12 numeric
                      -- dwarf address space is encoded in bitcode as +1; a value
-                     -- of zero means there is no dwarf address space present.
+                     -- of zero means there is no dwarf address space present:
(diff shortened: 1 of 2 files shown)


src/Data/LLVM/BitCode/IR/Metadata.hs
```diff
@@ -611,8 +611,16 @@ parseMetadataEntry vt mt pm (fromEntry -> Just r) =
...
         <*> (if length (recordFields r) <= 12
-             then pure Nothing
-             else Just                 <$> parseField r 12 numeric) -- didtDwarfAddressSpace
+             then pure Nothing  -- field not present
+             else do v <- parseField r 12 numeric
+                     -- dwarf address space is encoded in bitcode as +1; a value
+                     -- of zero means there is no dwarf address space present.
+                     -- The AST representation is the actual address space, or
+                     -- Nothing if there is no address space (indistinguishable
+                     -- from "field not present" for LLVM 4 and earlier).
```
RyanGlScott, line 620: Worth citing the place in the LLVM source code where this convention is described.

RyanGlScott (approved):

PR #255, opened 2023-09-26
Run fuzz testing CI on PR approval.

kquick:
These were actually sometimes failing, but they were being run on a cron cycle (nightly at 3AM) so notifications were only sent to the last person to modify the cron line (Aaron Tomb, whose github account is disabled... maybe because of this :-).

Additionally, this library is relatively stable (recent activity notwithstanding), so running these on a nightly basis is frivolous.  This changes them to run when a PR has been approved, ensuring master remains stable, but not inflicting the additional CI delay on every push.

### DIFF
--- a/.github/workflows/llvm-quick-fuzz.yml
+++ b/.github/workflows/llvm-quick-fuzz.yml
@@ -1,9 +1,20 @@
 name: llvm-quick-fuzz
 on:
-  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


RyanGlScott (approved):

kquick:
It worked!  Fuzz tests were run on approval.

PR #259, opened 2023-09-28
Use Monadic instead of Associative parsing of Metadata.

kquick:
Applicative construction is quadratic in performance, so this monadic parsing is faster (11 seconds faster on a 22MB bitcode file).

### DIFF
--- a/src/Data/LLVM/BitCode/IR/Metadata.hs
+++ b/src/Data/LLVM/BitCode/IR/Metadata.hs
@@ -424,6 +424,21 @@ parseMetadataEntry vt mt pm (fromEntry -> Just r) =
              fail $ unlines $ [ "Invalid record size: " ++ show len
                               , "Expected size of " ++ show lb ++ " or greater"
                               ] ++ msg
+
+      -- Helper for a common pattern which appears below in the parsing
+      ron n = do ctx <- getContext
+                 mdForwardRefOrNull ctx mt <$> parseField r n numeric
+
+  -- Note: the parsing cases below use a Monadic coding style, as opposed to an
+  -- Applicative style (as was originally used) for performance reasons:
+  -- Applicative record construction has quadratic size and corresponding
+  -- performance impacts (the initial conversion from Applicative to Monadic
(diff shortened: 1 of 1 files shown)


RyanGlScott (commented):
I'm rather surprised that this has as noticeable of a performance benefit as it does. I was aware that `Applicative` code could lead to [quadratic code _size_](https://gitlab.haskell.org/ghc/ghc/-/issues/10980), but I've never seen it lead to quadratic _runtime performance_ before. I can't help but wonder if the actual culprit is a difference in how GHC optimizes the `Applicative`-based code versus the `Monad`ic code, as GHC usually optimizes them to the same Core.

That being said, I'm keenly aware of how fickle GHC's optimizer can be, and I don't particularly feel like trying to minimize this issue. I'm happy with this approach, but I would appreciate a Note somewhere that explains that the use of monadic code was a deliberate choice in the name of performance. Perhaps something like `Note [Performance tricks]`, which could mention this and the use of unboxed types added in https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/197.

src/Data/LLVM/BitCode/IR/Metadata.hs
```diff
@@ -440,6 +424,21 @@ parseMetadataEntry vt mt pm (fromEntry -> Just r) =
...
+                 mdForwardRefOrNull ctx mt <$> parseField r n numeric
+
+  -- Note: the parsing cases below use a Monadic coding style, as opposed to an
+  -- Applicative style (as was originally used) for performance reasons:
+  -- Applicative record construction has quadratic size and corresponding
+  -- performance impacts (the initial conversion from Applicative to Monadic
+  -- saved 11s when parsing a 22MB bitcode file).
+  --
+  -- Additionally, this module uses RecordWildcards... a pragma that is not
+  -- normally adviseable but which does work to good effect in this situation to
```
RyanGlScott, line 439:
```suggestion
  -- normally advisable but which does work to good effect in this situation to
```

RyanGlScott (approved):

Issue #260, opened 2023-10-03
`llvm-pretty-bc-parser` inlines metadata nodes unnecessarily


RyanGlScott:
If you compile this program:

```c
// test.c
int main(void) {
  int x = 0;
  return x;
}
```

Like so:

```
$ clang-14 test.c -emit-llvm -g -c && clang-14 test.c -emit-llvm -g -S
```

Then this is what the resulting `test.ll` file will look like:

<details>

```ll
; ModuleID = 'test.c'
source_filename = "test.c"
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
target triple = "x86_64-pc-linux-gnu"

; Function Attrs: noinline nounwind optnone uwtable
define dso_local i32 @main() #0 !dbg !10 {
  %1 = alloca i32, align 4
  %2 = alloca i32, align 4
  store i32 0, i32* %1, align 4
  call void @llvm.dbg.declare(metadata i32* %2, metadata !15, metadata !DIExpression()), !dbg !16
  store i32 0, i32* %2, align 4, !dbg !16
  %3 = load i32, i32* %2, align 4, !dbg !17
  ret i32 %3, !dbg !18
}

; Function Attrs: nofree nosync nounwind readnone speculatable willreturn
declare void @llvm.dbg.declare(metadata, metadata, metadata) #1

attributes #0 = { noinline nounwind optnone uwtable "frame-pointer"="all" "min-legal-vector-width"="0" "no-trapping-math"="true" "stack-protector-buffer-size"="8" "target-cpu"="x86-64" "target-features"="+cx8,+fxsr,+mmx,+sse,+sse2,+x87" "tune-cpu"="generic" }
attributes #1 = { nofree nosync nounwind readnone speculatable willreturn }

!llvm.dbg.cu = !{!0}
!llvm.module.flags = !{!2, !3, !4, !5, !6, !7, !8}
!llvm.ident = !{!9}

!0 = distinct !DICompileUnit(language: DW_LANG_C99, file: !1, producer: "Ubuntu clang version 14.0.0-1ubuntu1.1", isOptimized: false, runtimeVersion: 0, emissionKind: FullDebug, splitDebugInlining: false, nameTableKind: None)
[...]

PR #265, opened 2024-01-02
Adds a new C++ test, with associated known_bugs entry for current failure

kquick:
[Requires #264 to be actually used.]

### DIFF
--- a/disasm-test/known_bugs/p2.cpp
+++ b/disasm-test/known_bugs/p2.cpp
@@ -1,0 +1,10 @@
+##> rootMatchName: p2.cpp
+##> summary: failure to llvm-as the llvm-disasm results
+
+Using this module's llvm-disasm to convert the clang++ generated bitcode into assembly (.ll), the resulting assembly is invalid for llvm-as:
+
+    error: expected type
+    define linkonce_ddr default i32 @_ZN4uorb14uorb_advertiseEPK21orb_metadata(%class.uorb* %32
+
+This is demonstrated with llvm version 7, 11, 13, and 15: it may or may not occur
+with other LLVM versions.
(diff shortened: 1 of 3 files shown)


langston-barrett (approved):

disasm-test/known_bugs/p2.cpp
```diff
@@ -0,0 +1,10 @@
+##> rootMatchName: p2.cpp
```
langston-barrett, line 1: Is there a corresponding issue number for this bug? Should we make one?

disasm-test/known_bugs/p2.cpp
```diff
@@ -0,0 +1,10 @@
...
+##> summary: failure to llvm-as the llvm-disasm results
+
+Using this module's llvm-disasm to convert the clang++ generated bitcode into assembly (.ll), the resulting assembly is invalid for llvm-as:
+
+    error: expected type
+    define linkonce_ddr default i32 @_ZN4uorb14uorb_advertiseEPK21orb_metadata(%class.uorb* %32
```
peterohanley, line 7: This error is just a typo, it should be linkonce_odr. The data member is named correctly, the error is here: https://github.com/elliottt/llvm-pretty/blob/8124fc0265b6ca5fde5812c789f40b2ea54af678/src/Text/LLVM/PP.hs#L463

With this fixed, there is an instance of #222 which is probably the same cause.

peterohanley: I reduced p2.cpp (for #222) to
```
struct h {
  bool a;
};

void g(struct h *s) {
}

void f() {
  struct h s;
}
```

I do not have a fix yet, but the bitcode is now small enough to walk through manually.

RyanGlScott: Great sleuthing!

I actually think the culprit here isn't #222, but rather #260, as the diff failure is caused by improper inlining. I'll add this example to the comments in #260.

PR #270, opened 2024-02-22
Support GHC 9.8, drop GHC 9.2

langston-barrett:


RyanGlScott:
The test suite likely won't compile without first resolving https://github.com/kquick/kvitable/issues/2.

langston-barrett:
Hmm, it looks like the diagnostics in that issue only consist of warnings, perhaps we can just bump the `base` bound and ignore the warnings until a more satisfying resolution is found?

cabal.project
```diff
@@ -1,3 +1,5 @@
 packages: .
 optional-packages: llvm-pretty
 constraints: directory >= 1.2.7
+-- See https://github.com/kquick/kvitable/issues/2
+allow-newer: kvitable:base
```
langston-barrett, line 5: I think the plan is now to merge https://github.com/kquick/kvitable/pull/4 and upload a new minor version of kvitable to Hackage. We can then delete these lines.

kquick: Version 1.0.3.0 of kvitable is now available from Hackage.

kquick (changes requested):
The `nix-ci.yml` also needs to be updated to add GHC 9.8, which will necessitate `flake.nix` changes (essentially you can duplicate the changes in https://github.com/kquick/kvitable/commit/4b4043b9bddfe67ae2ea767e18557908b861f154).

langston-barrett:
At https://github.com/GaloisInc/llvm-pretty-bc-parser/commit/b85a510747039e83e02db1a12310d0ec6751d6ed#r139131670, @kquick said:

> What do you think of leaving the older GHC's, but marking them (via comment) as "optimistic builds; no longer supported"? My thinking here is that (a) unlike using regular ubuntu packages, the locking provided by flakes makes it easy to access these older versions, (b) these builds are pretty fast and lowe cost, and (c) this can let us explicitly know when our trailing horizon is updated.

I don't have a strong stance here, if we do go forward with this we should update `doc/development.md` and the README to reflect the new policy.

kquick:
We wouldn't need to change the docs because we would still have the same policy of committing to support the recent 3-version window.  This would just be adding the older GHC builds into the nix ci to inform us regarding the trailing support horizon.

kquick (approved):

PR #275, opened 2024-08-28
Prepare release 0.4.2.0

mccleeary-galois:
Updated Changelog and submodule bumps for llvm-pretty

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Revision history for llvm-pretty-bc-parser
 
+## 0.4.2.0 (August 2024)
+
+* Add support for GHC 9.8 and drop official support of 9.2.
+
+* Add support for new atomic operations in LLVM 9+.
+
 ## 0.4.1.0 (January 2024)
 
 * Add preliminary support for LLVM versions up through 16.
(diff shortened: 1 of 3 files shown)


llvm-pretty
RyanGlScott, line 1: Now that we've bumped the `llvm-pretty` submodule to `0.12.1.0`, we should bump the lower version bounds in `llvm-pretty-bc-parser.cabal` so that it requires `llvm-pretty >= 0.12.1.0`. Because the current version of `llvm-pretty-bc-parser` references the new `AtomicRWOp` data constructors, it won't build with any older version of `llvm-pretty`.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Revision history for llvm-pretty-bc-parser
 
+## 0.4.2.0 (August 2024)
+
+* Added support for GHC 9.8 and dropped official support of 9.2
+
+* Added support for new atomic operations in LLVM 9+.
```
RyanGlScott, line 7: A minor thing: all of the previous changelog entries use present tense, so we may as well stick with that convention:

```suggestion
* Add support for GHC 9.8 and drop official support of 9.2.

* Add support for new atomic operations in LLVM 9+.
```

RyanGlScott (approved):

PR #280, opened 2024-11-26
Add support for LLVM address spaces

peterohanley:
Address spaces default to 0 and are currently mostly used by CHERI for capabilities (address space 200).

The thread_local property is properly read and reported.

Various formatting changes are implemented to more closely match llvm-dis output.

### DIFF
--- a/src/Data/LLVM/BitCode/IR/Function.hs
+++ b/src/Data/LLVM/BitCode/IR/Function.hs
@@ -83,7 +83,7 @@ parseAlias r = do
   (name, offset) <- oldOrStrtabName n r
   let field i = parseField r (i + offset)
   ty       <- getType =<< field 0 numeric
-  _addrSp  <-             field 1 unsigned
+  addrSp <- AddrSpace <$> field 1 numeric
   tgt      <-             field 2 numeric
   lnk      <-             field 3 linkage
   vis      <-             field 4 visibility
@@ -91,7 +91,7 @@ parseAlias r = do
 
   -- XXX: is it the case that the alias type will always be a pointer to the
   -- aliasee?
-  _   <- pushValue (Typed (PtrTo ty) (ValSymbol name))
+  _   <- pushValue (Typed (PtrTo addrSp ty) (ValSymbol name))
 
   return PartialAlias
     { paLinkage    = Just lnk
@@ -122,15 +122,18 @@ type DeclareList = Seq.Seq FunProto
 -- | Turn a function prototype into a declaration.
 finalizeDeclare :: FunProto -> Parse Declare
 finalizeDeclare fp = case protoType fp of
-  PtrTo (FunTy ret args va) -> return Declare
+  PtrTo adr (FunTy ret args va) -> return Declare
     { decLinkage    = protoLinkage fp
     , decVisibility = protoVisibility fp
+    , decUnnamedAddr = protoUnnamedAddr fp
     , decRetType    = ret
     , decName       = protoSym fp
     , decArgs       = args
     , decVarArgs    = va
     , decAttrs      = []
     , decComdat     = protoComdat fp
+    -- TODO what if this isn't equal to adr?
+    , decAddrSpace  = protoAddrSpace fp
     }
   _ -> fail "invalid type on function prototype"
 
@@ -144,6 +147,7 @@ type DefineList = Seq.Seq PartialDefine
 data PartialDefine = PartialDefine
   { partialLinkage    :: Maybe Linkage
   , partialVisibility :: Maybe Visibility
+  , partialAddrSpace  :: AddrSpace
   , partialGC         :: Maybe GC
   , partialSection    :: Maybe String
   , partialRetType    :: Type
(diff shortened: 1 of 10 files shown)


src/Data/LLVM/BitCode/IR/Function.hs
```diff
@@ -122,15 +122,18 @@ type DeclareList = Seq.Seq FunProto
...
     , decVisibility = protoVisibility fp
+    , decUnnamedAddr = protoUnnamedAddr fp
     , decRetType    = ret
     , decName       = protoSym fp
     , decArgs       = args
     , decVarArgs    = va
     , decAttrs      = []
     , decComdat     = protoComdat fp
+    -- TODO what if this isn't equal to adr?
+    , decAddrSpace  = protoAddrSpace fp
```
peterohanley, line 136: I think this would still be a sensible construct if they were not equal, but it hasn't been used.

peterohanley:
The formatting changes to match llvm-dis break golden tests. They're meant for helping compare output, so I'll remove them for now and just leave the semantic changes.

src/Data/LLVM/BitCode/IR/Function.hs
RyanGlScott: Hm, that is tricky. Do you know what LLVM does here? If nothing else, it might be worth checking for an (opt-in) `assert`.

src/Data/LLVM/BitCode/IR/Globals.hs
```diff
@@ -50,18 +50,37 @@ parseGlobalVar n r = label "GLOBALVAR" $ do
   vis <- if length (recordFields r) > (6 + offset) && not (link `elem` [Internal, Private])
                 then field 6 visibility
                 else pure DefaultVisibility
+  tl <- if length (recordFields r) > (7 + offset) -- && not (link `elem` [Internal, Private])
```
RyanGlScott, line 53: Can this comment be deleted? I think thread locality is parsed independently of the linkage, based on my reading of the LLVM implementation [here](https://github.com/llvm/llvm-project/blob/803bd812b16d1454b8fc0d6d66d7da793408b34d/llvm/lib/Bitcode/Reader/BitcodeReader.cpp#L4086-L4088).

src/Data/LLVM/BitCode/IR/Globals.hs
```diff
@@ -50,18 +50,37 @@ parseGlobalVar n r = label "GLOBALVAR" $ do
...
+    then do
+      field 8 unnamedAddr
+    else return Nothing
+
+  addrspace <- if explicitTy
+                  then return . AddrSpace . fromIntegral $ shiftR mask 2
+                  else case ptrty of
+                         PtrTo as _ -> return as
+                         PtrOpaque as -> return as
+                         _ -> fail $ "Invalid type for value: " ++ show ptrty
```
RyanGlScott, line 68: Can this `case` expression be simplified by defining it in terms of `ptrAddrSpace` from https://github.com/GaloisInc/llvm-pretty/pull/148?

src/Data/LLVM/BitCode/Record.hs
```diff
@@ -182,5 +182,10 @@ oldOrStrtabName n r = do
           Just st -> do
             offset <- parseField r 0 numeric
             len <- parseField r 1 numeric
-            return (resolveStrtabSymbol st offset len, 2)
+            if len == 0
+            then do
+              n <- show <$> nextSymbolId
+              return (Symbol n, 2)
```
RyanGlScott, line 188: Can you leave some comments explaining what is going on in the `len == 0` case? From [this commit](https://github.com/GaloisInc/llvm-pretty-bc-parser/commit/c56e3da2124fddc1c249a765c6cb304299bcf8b0), I gather that it related to unnamed symbols in some way, but the details aren't clear to me.

src/Data/LLVM/BitCode/IR/Types.hs
```diff
@@ -116,11 +116,13 @@ parseTypeBlockEntry (fromEntry -> Just r) = case recordCode r of
...
-    when (length (recordFields r) == 2) $ do
-      -- We do not currently store address spaces in the @llvm-pretty@ AST.
-      _space <- field 1 keep
-      return ()
-    addType (PtrTo ty)
+    space <- AddrSpace <$> if length (recordFields r) == 2
+      then
+        field 1 numeric
+      else
+        return 0
```
kquick, line 123: Another candidate for defaultAddrSpace, alhtough the `AddrSpace` constructor will need to move to the `then`.

src/Data/LLVM/BitCode/Parse.hs
```diff
@@ -173,6 +185,22 @@ setModVersion v = Parse $ do
 getModVersion :: Parse Int
 getModVersion = Parse (psModVersion <$> get)
 
+setDataLayout :: DataLayout -> Parse ()
+setDataLayout v = Parse $ do
+  ps <- get
+  put $! ps { psDataLayout = Just v }
```
kquick, line 191:
```suggestion
 setDataLayout v = Parse $ modify $ \ps -> ps { psDataLayout = Just v }
```

PR #281, opened 2025-02-03
Fix file-local symtab lookups

kquick:
An LLVM Bitcode file, it is comprised of nested Blocks of information, with
records available at each block.  Different blocks hold different types of
information, and the nesting represents program scope and what is
defined/accessible at a particular point within the program.

The `Parse` and `Finalize` monads maintain a set of symbol tables to use for
lookups, as stored in the `Env` structure referenced by those monads:

  * `valSymtab` for value references

  * `fnSymtab` for function references

  * `bbSymTab` for basic-block references, either "named" by association with
    a symbol name (like the function's entry block) or "anonymous", where
    there is no direct symbol but there may be a block label (like for a goto
    statement or similar surface language construct.)

These lookups are necessary to "relabel" values that reference symbol or label
addresses in the code with the actual targets as resolved by processing the
entirety of the LLVM bitcode file.  The llvm-pretty-bc-parser runs in two
phases: the initial phase where it processes the LLVM bitcode stream to create
"Partial" representations of all of the elements, followed by a "finalization"
phase where it performs all of the label references above (via the llvm-pretty
`relabel` operation), as well as other fixups to convert the "Partial" data
structures into the structures defined by the `llvm-pretty` AST.

However, the tables in the `Env` structure previously only held the direct
symbol tables (i.e. `VALUE_SYMTAB_BLOCK`, found immediately within the
module's top-level `MODULE_BLOCK`.  Symbol resolution was performed by calling
`bbEntryName` or `requireBbEntryName`.  This was sufficient to resolve global
symbol references.

When finalizing functions, the Function finalization defined an alternate
resolution type (`BlockLookup`) and function (`lookupBlockName`) to use with
`relabel` for handling the code within the function.  The `bbEntryName` and
`requireBbEntryName` did not accept a symbol (identifying the function or
`Define` in LLVM IR parlance) so they could only reference the top-level
symbol tables, whereas `lookupBlockName` was used if the `relabel` indicated
there was an associated symbol: it would find the (partial) symbol table for
the function (`Define`) associated with that symbol and perform the `bbSymtab`
lookup from that block.

This duplication of functionality should have been a suggestion that something
was not right, particularly when the `bbEntryName` and `requireBbEntryName`
uses simply discarded any supplied symbol name.  But it worked most of the
time, so this went unnoticed for quite some time (eleven years or so), but it
finally broke when the assumptions about the ability to discard symbol names
was proven to be a bad.

What was it that revealed the invalidity of this assumption?

Answer: Function-local label references.

Consider the following

```
  int has_local_static(int x) {
    static const void *const disptab[NUM] = { &&fn1, &&fn2, &&fn3 };
    int y;
  fn1:
    ...
    goto *disptab[y];
  fn2:
    ...
    goto *disptab[y];
  fn3:
    return y;
  }
```

In this code, `disptab` is a local static.  To implement this, clang adds the
definition of `disptab` to the top-level `VALUE_SYMTAB_BLOCK`.  However, the
values in that list are local references to `has_local_static`.  The `relabel`
operation will supply the `has_local_static` symbol, but VALUE_SYMTAB_BLOCK
finalization (`finalizeGlobal`) uses `requireBbEntryName` and discards that
supplied symbol, thus failing to find the symbol in the top-level symbol block
and halting the parse with `fail`--or even worse, finding an unrelated global
basic-block reference and using that to create invalid references.  Neither of
these is any kind of good.

To fix this, this PR enhances the reader environment for Finalizeto also maintain the `Define` symtabs entries obtained during parsing, and then enhancing `bbEntryName` (and
`requireBbEntryName`) to lookup any symbols and perform `bbSymtab` resolution in the symbol
table associated with that definition.  This means the `relabel`-supplied symbol
names are no longer ignored, and also means that `bbEntryName` is capable of
performing the full resolution that `lookupBlockName` was concocted to handle,
so the latter is removed as obsolete.  These changes allow us to successfully
parse the bitcode for the `has_local_static` function above (also added as a
test in `disasm-test/tests/localstatic.c`).

A test is included; running the `cabal run disasm-test` or simply `cabal run llvm-disasm -- disasm_test/test/localstatic.bc` at the commit where this was introduced shows the error; subsequent commits fix the issue.  Unfortunately, the test still fails round-trip as described in the known_bugs file added; this is for a somewhat separate issue that I would prefer to fix independently.

All commits are atomic; this PR can be split up if desired.

### DIFF
--- a/disasm-test/known_bugs/llvm_dbg_declare_inline
+++ b/disasm-test/known_bugs/llvm_dbg_declare_inline
@@ -2,6 +2,8 @@
 ##> rootMatchName: localstatic.ll
 ##> summary: llvm.dbg.declare should use md refs and not inline
 
+Logged in https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/260
+
 Attempts to inline declare the !DILocalVariable for the file-local static:
 
   call void @llvm.dbg.declare(metadata i32* %2, metadata !DILocalVariable(scope: !DISubprogram(scope: !3, name: "has_local_static", file: !3, line: 5, type: !9, isLocal: false, isDefinition: true, scopeLine: 5, virtuality: 0, virtualIndex: 0, flags: 256, isOptimized: false, unit: !2, retainedNodes: !4), name: "x", file: !DIFile(filename: "localstatic.c", directory: "/home/kquick/work/DFAMS/tp241209/sources/llvm-pretty-bc-parser/disasm-test/tests"), line: 5, type: !DIBasicType(tag: 36, name: "int", size: 32, align: 0, encoding: 5, flags: 0), arg: 1, flags: 0, align: 0), metadata !DIExpression()), !dbg !DILocation(line: 5, column: 26, scope: !0)
--- a/disasm-test/known_bugs/llvm_dbg_declare_inline
+++ b/disasm-test/known_bugs/llvm_dbg_declare_inline
@@ -1,0 +1,30 @@
+##> rootMatchName: localstatic.c
+##> rootMatchName: localstatic.ll
+##> summary: llvm.dbg.declare should use md refs and not inline
+
+Attempts to inline declare the !DILocalVariable for the file-local static:
+
+  call void @llvm.dbg.declare(metadata i32* %2, metadata !DILocalVariable(scope: !DISubprogram(scope: !3, name: "has_local_static", file: !3, line: 5, type: !9, isLocal: false, isDefinition: true, scopeLine: 5, virtuality: 0, virtualIndex: 0, flags: 256, isOptimized: false, unit: !2, retainedNodes: !4), name: "x", file: !DIFile(filename: "localstatic.c", directory: "/home/kquick/work/DFAMS/tp241209/sources/llvm-pretty-bc-parser/disasm-test/tests"), line: 5, type: !DIBasicType(tag: 36, name: "int", size: 32, align: 0, encoding: 5, flags: 0), arg: 1, flags: 0, align: 0), metadata !DIExpression()), !dbg !DILocation(line: 5, column: 26, scope: !0)
+
+, but llvm-as rejects that:
+
+  llvm-as: file.ll:L:C: error: missing 'distinct', required for !DISubprogram that is a Definition
+
+Adding 'distinct' in various places on the line does not work
+
+The llvm-dis handles this as:
+
+  call void @llvm.dbg.declare(metadata i32* %2, metadata !21, metadata !DIExpression()), !dbg !22
+
+where:
+
+!2 = distinct !DISubprogram(name: "has_local_static", scope: !3, file: !3, line: 5, type: !4, scopeLine: 5, flags: DIFlagPrototyped, spFlags: DISPFlagDefinition, unit: !7, retainedNodes: !8)
+!21 = !DILocalVariable(name: "x", arg: 1, scope: !2, file: !3, line: 5, type: !6)
+
+Note that for the llvm-disasm version, we do have:
+
+!0 = distinct !DISubprogram(scope: !3, name: "has_local_static", file: !3, line: 5, type: !9, isLocal: false, isDefinition: true, scopeLine: 5, virtuality: 0, virtualIndex: 0, flags: 256, isOptimized: false, unit: !2, retainedNodes: !4)
+!21 = !DILocalVariable(scope: !0, name: "x", file: !3, line: 5, type: !7, arg: 1, flags: 0, align: 0)
+
+So apparently it is mostly needed to have the llvm.dbg.declare use the UnnamedMd
+indices (if available) rather than inlining them.
--- a/src/Data/LLVM/BitCode/IR/Values.hs
+++ b/src/Data/LLVM/BitCode/IR/Values.hs
@@ -52,10 +52,7 @@ getValue :: ValueTable -> Type -> Int -> Parse (Typed PValue)
 getValue vt ty n = label "getValue" (getFnValueById' (Just vt) ty =<< adjustId n)
 
 -- | Lookup a value by its absolute id, or perhaps some metadata.
-getFnValueById' :: HasMdTable m
-                => HasParseEnv m
-                => HasValueTable m
-                => MonadFail m
+getFnValueById' :: (HasMdTable m, HasParseEnv m, HasValueTable m, MonadFail m)
                 => Maybe ValueTable -> Type -> Int -> m (Typed PValue)
 getFnValueById' mbVt ty n = case ty of
 
--- a/src/Data/LLVM/BitCode/IR/Values.hs
+++ b/src/Data/LLVM/BitCode/IR/Values.hs
@@ -44,15 +44,20 @@ getValueTypePair t r ix = do
 --getValueNoFwdRef :: Type -> Int -> Parse (Typed PValue)
 --getValueNoFwdRef ty n = label "getValueNoFwdRef" (getFnValueById ty =<< adjustId n)
 
-getFnValueById :: Type -> Int -> Parse (Typed PValue)
+getFnValueById :: (HasMdTable m, HasParseEnv m, HasValueTable m, MonadFail m)
+               => Type -> Int -> m (Typed PValue)
 getFnValueById  = getFnValueById' Nothing
 
 getValue :: ValueTable -> Type -> Int -> Parse (Typed PValue)
 getValue vt ty n = label "getValue" (getFnValueById' (Just vt) ty =<< adjustId n)
 
 -- | Lookup a value by its absolute id, or perhaps some metadata.
-getFnValueById' :: Maybe ValueTable -> Type -> Int -> Parse (Typed PValue)
-getFnValueById' mbVt ty n = label "getFnValueById'" $ case ty of
+getFnValueById' :: HasMdTable m
+                => HasParseEnv m
+                => HasValueTable m
+                => MonadFail m
+                => Maybe ValueTable -> Type -> Int -> m (Typed PValue)
+getFnValueById' mbVt ty n = case ty of
 
   PrimType Metadata -> do
     cxt <- getContext
--- a/src/Data/LLVM/BitCode/Parse.hs
+++ b/src/Data/LLVM/BitCode/Parse.hs
@@ -20,8 +20,9 @@ import qualified Control.Monad.Fail -- makes fail visible for instance
 #endif
 import           Control.Monad.Fix (MonadFix)
 import           Control.Monad.Except (MonadError(..), Except, runExcept)
-import           Control.Monad.Reader (MonadReader(..), ReaderT(..))
-import           Control.Monad.State.Strict (MonadState(..), StateT(..))
+import           Control.Monad.Reader (MonadReader(..), ReaderT(..), asks)
+import           Control.Monad.State.Strict (MonadState(..), StateT(..)
+                                            , gets, modify)
 import           Data.Maybe (fromMaybe)
 import           Data.Semigroup
 import           Data.Typeable (Typeable)
@@ -55,7 +56,11 @@ formatError err
 
 newtype Parse a = Parse
   { unParse :: ReaderT Env (StateT ParseState (Except Error)) a
-  } deriving (Functor, Applicative, MonadFix)
+  } deriving (Functor, Applicative, MonadFix
+             , MonadReader Env
+             , MonadState ParseState
+             , MonadError Error
+             )
 
 instance Monad Parse where
 #if !MIN_VERSION_base(4,11,0)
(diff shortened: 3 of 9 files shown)


disasm-test/known_bugs/llvm_dbg_declare_inline
```diff
@@ -0,0 +1,30 @@
...
+!2 = distinct !DISubprogram(name: "has_local_static", scope: !3, file: !3, line: 5, type: !4, scopeLine: 5, flags: DIFlagPrototyped, spFlags: DISPFlagDefinition, unit: !7, retainedNodes: !8)
+!21 = !DILocalVariable(name: "x", arg: 1, scope: !2, file: !3, line: 5, type: !6)
+
+Note that for the llvm-disasm version, we do have:
+
+!0 = distinct !DISubprogram(scope: !3, name: "has_local_static", file: !3, line: 5, type: !9, isLocal: false, isDefinition: true, scopeLine: 5, virtuality: 0, virtualIndex: 0, flags: 256, isOptimized: false, unit: !2, retainedNodes: !4)
+!21 = !DILocalVariable(scope: !0, name: "x", file: !3, line: 5, type: !7, arg: 1, flags: 0, align: 0)
+
+So apparently it is mostly needed to have the llvm.dbg.declare use the UnnamedMd
+indices (if available) rather than inlining them.
```
RyanGlScott, line 30: Is this the same issue as described in https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/260? If so, it would be worth citing this issue.

src/Data/LLVM/BitCode/Parse.hs
```diff
@@ -55,7 +56,11 @@ formatError err
 
 newtype Parse a = Parse
   { unParse :: ReaderT Env (StateT ParseState (Except Error)) a
-  } deriving (Functor, Applicative, MonadFix)
+  } deriving (Functor, Applicative, MonadFix
+             , MonadReader Env
+             , MonadState ParseState
+             , MonadError Error
+             )
```
RyanGlScott, line 63: Minor whitespace formatting suggestion:

```suggestion
  } deriving ( Functor, Applicative, MonadFix
             , MonadReader Env
             , MonadState ParseState
             , MonadError Error
             )
```

src/Data/LLVM/BitCode/IR/Values.hs
```diff
@@ -44,15 +44,20 @@ getValueTypePair t r ix = do
...
 getValue vt ty n = label "getValue" (getFnValueById' (Just vt) ty =<< adjustId n)
 
 -- | Lookup a value by its absolute id, or perhaps some metadata.
-getFnValueById' :: Maybe ValueTable -> Type -> Int -> Parse (Typed PValue)
-getFnValueById' mbVt ty n = label "getFnValueById'" $ case ty of
+getFnValueById' :: HasMdTable m
+                => HasParseEnv m
+                => HasValueTable m
+                => MonadFail m
+                => Maybe ValueTable -> Type -> Int -> m (Typed PValue)
```
RyanGlScott, line 59: The type signature for `getFnValueById` puts all of the constraints into a tuple:

```hs
getFnValueById :: (HasMdTable m, HasParseEnv m, HasValueTable m, MonadFail m)
               => ...
```

While the type signature for `getFnValueById'` puts each constraint on its own line:

```hs
getFnValueById' :: HasMdTable m
                => HasParseEnv m
                => HasValueTable m
                => MonadFail m
                => ...
```

It would be better to stick to a single convention, I think. I have a slight preference towards the former style, although I wouldn't push back if you advocate for the latter style.

src/Data/LLVM/BitCode/Parse.hs
```diff
@@ -153,10 +158,9 @@ setRelIds b = Parse $ do
   ps <- get
   put $! ps { psValueTable = (psValueTable ps) { valueRelIds = b }}
 
-getRelIds :: Parse Bool
-getRelIds  = do
-  ps <- Parse get
-  return (valueRelIds (psValueTable ps))
+getRelIds :: HasValueTable m => m Bool
+getRelIds  = valueRelIds <$> getValueTable
+  -- return (valueRelIds (psValueTable ps))
```
RyanGlScott, line 163: Can this commented-out line of code be removed?

src/Data/LLVM/BitCode/Parse.hs
```diff
@@ -574,6 +583,33 @@ getTypeId n = do
...
+-- label addresses in the code with the actual targets as resolved by
+-- processing the entirety of the LLVM bitcode file.  The
+-- llvm-pretty-bc-parser runs in two phases: the initial phase where it
+-- processes the LLVM bitcode stream to create "Partial" representations of
+-- all of the elements, followed by a "finalization" phase where it performs
+-- all of the label references above (via the llvm-pretty `relabel`
+-- operation), as well as other fixups to convert the "Partial" data
+-- structures into the structures defined by the `llvm-pretty` AST.
+
 type SymName = Either String Int
```
RyanGlScott, line 613: Currently, these Haddocks are describing the `type SymName = Either String Int` type synonym, which doesn't seem like what you intended. Should these Haddocks describe a different definition? Perhaps this was meant to be a Haddock named section (e.g., `$llvmBitcodeFile`) in the export list? Something else?

src/Data/LLVM/BitCode/Parse.hs
```diff
@@ -624,22 +660,22 @@ addFwdFNEntry :: Int -> Int -> ValueSymtab -> ValueSymtab
...
 entryNameMb n = do
-  symtab <- liftFinalize getValueSymtab
+  symtab <- getValueSymtab
   return $! fmap renderName
          $  IntMap.lookup n (valSymtab symtab) `mplus`
             IntMap.lookup n (fnSymtab symtab)
 
 -- | Lookup the name of an entry.
-entryName :: Int -> Parse String
+entryName :: HasParseEnv m => HasValueTable m => MonadFail m => Int -> m String
```
RyanGlScott, line 671: Similarly, I have a slight preference towards tupling the constraints here.

src/Data/LLVM/BitCode/Parse.hs
```diff
@@ -736,9 +772,14 @@ resolveStrtabSymbol (Strtab bs) start len =
 
 -- Finalize Monad --------------------------------------------------------------
 
+data FinalizeEnv = FinalizeEnv { parsedEnv :: Env
+                               , parsedMdTable :: ValueTable
+                               , parsedValueTable :: ValueTable
+                               }
```
RyanGlScott: Please add some Haddocks to this `FinalizeEnv` data type.

Also, it strikes me that we now have multiple `*Env` data types: we have `FinalizeEnv` here, and we also have a plain `Env` data type. This feels rather confusing. I wonder if `Env` should be renamed to `ParseEnv` to better distinguish the two?

src/Data/LLVM/BitCode/Parse.hs
```diff
@@ -497,9 +505,14 @@ extendSymtab symtab env = env { envSymtab = envSymtab env `mappend` symtab }
 addLabel :: String -> Env -> Env
 addLabel l env = env { envContext = l : envContext env }
 
-getContext :: Parse [String]
-getContext  = Parse (envContext `fmap` ask)
+class Monad m => HasParseEnv m where
+  getContext :: m [String]
```
RyanGlScott, line 509: Please add some Haddocks for this method—what does "context" mean here?

src/Data/LLVM/BitCode/Parse.hs
```diff
@@ -497,9 +505,14 @@ extendSymtab symtab env = env { envSymtab = envSymtab env `mappend` symtab }
...
+  -- | Retrieve the value symbol table
+  getValueSymtab :: m ValueSymtab
```
RyanGlScott, line 511: I can't help but wonder if it would be better to replace these two methods with a single method that returns an `Env` (or `ParseEnv`, or whatever we end up calling it).

src/Data/LLVM/BitCode/Parse.hs
```diff
@@ -833,14 +845,15 @@ failWithContext' msg =
        , errContext = envContext env
        }
 
-liftFinalize :: Finalize a -> Parse a
-liftFinalize (Finalize m) =
+liftFinalize :: FuncSymTabs -> Finalize a -> Parse a
```
RyanGlScott, line 848: Please leave some comments describing this `FuncSymTabs`. Under what circumstances should this be passed an empty `Map` versus a non-empty one?

RyanGlScott (approved):
Thanks! In the future, it would be helpful to file an issue first describing the (rather lengthy and nuanced) problem before filing a PR. This would be especially helpful since (AFAIU) this PR doesn't completely fix the underlying problem due to the roundtrip errors—that way, we could associate one issue with multiple PRs.

> A test is included; running the `cabal run disasm-test` or simply `cabal run llvm-disasm -- disasm_test/test/localstatic.bc` at the commit where this was introduced shows the error; subsequent commits fix the issue. Unfortunately, the test still fails round-trip as described in the known_bugs file added; this is for a somewhat separate issue that I would prefer to fix independently.

Yes, I would greatly appreciate if you could help with this. I strongly suspect that this is due to #260, an issue which is currently blocking support for LLVM 17 and later (see https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/261).

src/Data/LLVM/BitCode/Parse.hs
kquick: I've added some general Haddocks.

Passing an empty map is only done in `Function.hs` in two places. Both of those places I find a little suspicious. Resolving those suspicions would take an hour or two of investigation, code changes, and testing, whereas passing an empty map preserves the functionality that was originally there.  I'd rather not add documentation that asserts something I'm not confident about because I wouldn't want to mislead future development efforts.

src/Data/LLVM/BitCode/Parse.hs
kquick: They were originally distinct functions, but only defined for use in the `Parse` monad.  I needed them useable for the `Finalize` monad as well, but felt like the `Env` (and `FinalizeEnv`) were "internal implementation details" for these monads behind the API of these two functions.  

I did end up exposing the implementation detail just a little bit in the name of the constraint. The class could be modified to return the `Env` instead, but then these would again become distinct functions with the `HasParseEnv` constraint on them, which would need to still be present for their callers as well, so it only changes the "internal" implementation a little bit (towards increased complexity, IMHO).  To hide the `Env` aspect, I could change the  class name to something else but that name would be less directly useful, or split this into two classes with names associated with the single method contained in each, which would require adding extra constraints in various places.

Let me know if you feel strongly about renaming or splitting the class.

src/Data/LLVM/BitCode/Parse.hs
kquick: Added.

src/Data/LLVM/BitCode/Parse.hs
kquick: Haddocks added.

The `Env` could be renamed, but the two names are distinct, and `FinalizeEnv` is really only referenced and used "internally" in that small portion of the file that small portion of the file that implements the `Finalize` monad, so I was choosing in favor of not adding size to this PR by a renaming.

src/Data/LLVM/BitCode/Parse.hs
kquick: Tupled.

src/Data/LLVM/BitCode/IR/Values.hs
kquick: I have a slight preference for the stacked 1/line style because it makes commenting out individual constraints during dev easier.  That said, it's only a very slight preference, and I'm fine with tupling... so: done.

disasm-test/known_bugs/llvm_dbg_declare_inline
kquick: Due to github's insistence on leaving tracks, I believe you just did. :-)

kquick: And yes, I believe it is the same issue.

src/Data/LLVM/BitCode/Parse.hs
RyanGlScott: I don't feel strongly about it. Your rationale makes sense, so I'm fine with keeping the methods separate.

disasm-test/known_bugs/llvm_dbg_declare_inline
RyanGlScott: To be clear, I meant to cite #260 in this specific file.

RyanGlScott (approved):

sauclovian-g:
It's never too late to file an issue :-)

PR #289, opened 2025-03-18
Bump llvm-pretty submodule to current master

sauclovian-g:
Release prep.


marsella (commented):
Looks reasonable to me. Is there a blocker to also tagging a release for `llvm-pretty`?

marsella (approved):

sauclovian-g:
There is not, the idea was to do all the material updates beforehand in case one of them blew up.

PR #291, opened 2025-03-20
Prepare release 0.5.0

sauclovian-g:


llvm-pretty-bc-parser.cabal
```diff
@@ -75,7 +75,7 @@ Library
                        bytestring  >= 0.10,
                        containers  >= 0.4,
                        fgl         >= 5.5,
-                       llvm-pretty >= 0.12.1.0.99 && < 0.13,
+                       llvm-pretty >= 0.12.1.0.99 && < 0.14,
```
RyanGlScott, line 78: I think that we should bump the lower version bounds here to `0.13`. That way, we can upload this to Hackage without having development versions in the `.cabal` file.

(In a subsequent commit, these lower version bounds could then be raised to `0.13.0.0.99`.)

sauclovian-g: I assumed that was there for a reason so didn't want to change it.

Is there a reason to disallow the release version going forward? I wouldn't think so...

RyanGlScott: > Is there a reason to disallow the release version going forward?

We _could_ allow building with both `llvm-pretty-0.12.*` and `llvm-pretty-0.13.*`, but it would increase our maintenance burden to ensure that both major releases are supported. What's more, we are already vendoring `llvm-pretty` as a submodule, so in practice we are almost certainly only ever going to test one particular `llvm-pretty` release at a time. Making the lower version bounds reflect this, in my view, avoids a potential situation where we accidentally drop compatibility with `llvm-pretty-0.12.*` later while forgetting to bump the lower version bounds, which could lead to confusing build plan failures.

sauclovian-g: Fair enough, that's a good reason to drop 0.12. But I don't think there's any need to set the lower bounds to 0.13.0.0.99. Or is that needed in the nix case?

RyanGlScott: Ah, fair enough. We can wait until we are relying on features from an unreleased `llvm-pretty` version before bumping the lower version bounds beyond `0.13`.

RyanGlScott (approved):

llvm-pretty-bc-parser.cabal
kquick: It is not a nix thing.

The idea behind the a.b.c.d.99 (and followons of a.b.c.d.100, etc.) is that those are "dev versions" that allow downstream _dev_ packages to express lower-bound constraints to ensure recently developed features/changes that haven't been release yet are observed.  The explicit submodule will generally ensure the proper dependency is seen, but this is not always visible to all consumers, so enabling that 5th field for cabal-level dependency validation helps protect against that.

As an example, `llvm-pretty-bc-parser` explicitly references `llvm-pretty` via a direct submodule, so the dev version doesn't add any value.  But if package `foo` uses both `llvm-pretty-bc-parser` and `llvm-pretty` via direct submodules (which we generally do, rather than using recursive submodules), then if a `foo` developer updates `llvm-pretty-bc-parser` to a newer commit but forgets to update the `foo` reference to `llvm-pretty`, then the minimum version in llvm-pretty-bc-parser's cabal file will prevent the incorrect build.

It's a little bit of extra ceremony around the build, but having that 5th digit present guides regular developers into knowing how to address this situation.

sauclovian-g: I know what the 99 is for :-) sorry if that wasn't clear. And yeah, now I see why that particular constraint might be wanted...

PR #301, opened 2025-08-11
ci: Support GHC 9.12.2, drop 9.6.6

### LINKED ISSUE #284: Support GHC 9.12
- [ ] Allow the [new version of `base`](https://wiki.haskell.org/Base_package#Versions) in the Cabal `build-depends`, build and test
- [ ] Fix any new warnings from [`-Wdefault`](https://downloads.haskell.org/ghc/latest/docs/users_guide/using-warnings.html#ghc-flag-Wdefault)
- [ ] Add a new [`cabal freeze`](https://cabal.readthedocs.io/en/stable/cabal-commands.html#cabal-freeze) file, if the project uses them
- [ ] Add the new GHC version in CI configuration
- [ ] Add the new GHC version to the Cabal [`tested-with` field](https://cabal.readthedocs.io/en/stable/cabal-package-description-file.html#pkg-field-tested-with), if present
- [ ] Update any mentions of GHC versions in the documentation (e.g., README), if present
- [ ] Optionally repeat the above steps, but removing / deprecating an older GHC version

langston-barrett:
Fixes #284.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,7 @@ jobs:
       matrix:
         os: [ubuntu-22.04]
         # See doc/developing.md
-        ghc: ["9.6.6", "9.8.4", "9.10.1"]
+        ghc: ["9.8.4", "9.10.1", "9.12.2"]
     name: build - ${{ matrix.ghc }} - ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/llvm-quick-fuzz.yml
+++ b/.github/workflows/llvm-quick-fuzz.yml
@@ -34,7 +34,7 @@ jobs:
       matrix:
         os: [ubuntu-22.04]
         # See doc/developing.md
-        ghc: ["9.10.1"]
+        ghc: ["9.12.2"]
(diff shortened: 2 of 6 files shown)


src/Data/LLVM/BitCode/IR/Metadata.hs
```diff
@@ -316,7 +315,7 @@ data PartialUnnamedMd = PartialUnnamedMd
   { pumIndex    :: Int
   , pumValues   :: PValMd
   , pumDistinct :: Bool
-  } deriving (Data, Eq, Ord, Generic, Show, Typeable)
+  } deriving (Data, Eq, Ord, Generic, Show)
```
kquick: I'm hesitant at removing this and the associated Derivation pragmas: although we are only *committing* to support three GHC revisions, this is a pretty low level package so it should lead on version support but trail on version deprecation.

kquick: Per the conversation and verification in `llvm-pretty`, this does not break builds from as far back as GHC 8.10, so I retract my concern.

langston-barrett: Here's that conversation: https://github.com/GaloisInc/llvm-pretty/pull/165#discussion_r2658006220

langston-barrett:
@kquick Looks like it is now available: https://github.com/NixOS/nixpkgs/blob/08008e2e9cb08722df3eacb8697936e6c8bf82b3/pkgs/top-level/haskell-packages.nix#L293

Issue #302, opened 2025-08-26
Support parsing bitcode produced by Zig


RyanGlScott:
The [Zig](https://ziglang.org/) compiler produces somewhat unusually shaped LLVM bitcode as compared to Clang. This issue aims to document what steps we would need to perform in order to support Zig-generated bitcode properly.

Throughout this issue, I will be using the bitcode that Zig generated from this program:

```zig
// test.zig
export fn add(a: i32, b: i32) i32 {
    return a + b;
}
```

Compiled like so:

```
$ zig version
0.16.0-dev.27+83f773fc6
$ zig build-lib -femit-llvm-bc -OReleaseFast test.zig
```

This produces a `test.bc` bitcode file. Here are the issues (in order) that I encountered when loading this bitcode file into `llvm-pretty-bc-parser`:

## `match failed` [...] `TYPE_BLOCK`

The first issue I ran into is:

```
> parseBitCodeFromFileWithWarnings "test.bc" >>= \x -> case x of Left err -> putStrLn (formatError err); Right _ -> pure ()
match failed
from:
	TYPE_BLOCK
	type symbol table
	MODULE_BLOCK
	Bitstream
```

This ultimately arises from how the type table is parsed here:

https://github.com/GaloisInc/llvm-pretty-bc-parser/blob/122aa18fc052db1adfdb5456d67223d4e4590499/src/Data/LLVM/BitCode/IR/Types.hs#L50-L51

Where `numEntry` is defined here:

https://github.com/GaloisInc/llvm-pretty-bc-parser/blob/122aa18fc052db1adfdb5456d67223d4e4590499/src/Data/LLVM/BitCode/IR/Types.hs#L25-L27

`llvm-pretty-bc-parser` expects `TYPE_CODE_NUMENTRY` to live in an unabbreviated record, but Zig's compiler happens to put `TYPE_CODE_NUMENTRY` in an abbreviated record instead. Fair enough, I suppose—I'm not sure why `llvm-pretty-bc-parser` is so picky here. The following (untested) patch appears to fix that issue:

```diff
diff --git a/src/Data/LLVM/BitCode/IR/Types.hs b/src/Data/LLVM/BitCode/IR/Types.hs
index ef564bd..a5be591 100644
--- a/src/Data/LLVM/BitCode/IR/Types.hs
+++ b/src/Data/LLVM/BitCode/IR/Types.hs
@@ -24,7 +24,7 @@ import           Data.Ord (comparing)

[...]

PR #303, opened 2025-09-18
Update CST_CODE_STRING parsing for newer alternative format.

kquick:
Fixes parse error noted as

    CST_CODE_STRING
    CONSTANTS_BLOCK
    ...

### DIFF
--- a/src/Data/LLVM/BitCode/IR/Constants.hs
+++ b/src/Data/LLVM/BitCode/IR/Constants.hs
@@ -6,6 +6,7 @@
 
 module Data.LLVM.BitCode.IR.Constants where
 
+import           Control.Applicative ( (<|>) )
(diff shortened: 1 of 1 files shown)


RyanGlScott (approved):
Discussing this with @kquick, it is unfortunately not straightforward to distill a minimal example of this. In light of this, I am willing to accept this change.

Longer-term, I think we should reconsider how `llvm-pretty-bc-parser` handles operands that may or may not be represented with `FieldArray`s, as this PR demonstrates yet another instance where the difference can be fatal (see https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/302 for a prior occurrence). I've opened https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/304 for this.

PR #317, opened 2025-10-29
Support parsing flags in cast-related instructions (`ZExt`, `UiToFp`, and `Trunc`)

### LINKED ISSUE #306: Support flags in cast-related instructions (added in LLVM 18+)
LLVM 18.0 adds a `nneg` flag to the `zext` instruction (see commit https://github.com/llvm/llvm-project/commit/ed3f06b9b393cd51e78e5fbc7a46bce090c1817a). Per the [LLVM Language Reference](https://releases.llvm.org/18.1.0/docs/LangRef.html#zext-to-instruction):

> The `nneg` (non-negative) flag, if present, specifies that the operand is non-negative. This property may be used by optimization passes to later convert the `zext` into a `sext`.
>
> [...]
>
> If the `nneg` flag is set, and the `zext` argument is negative, the result is a poison value.

The `nneg` flag is stored as an extra field in the bitcode record, which causes `llvm-pretty-bc-parser` to crash when it tries to parse it. This can be seen in the following test case, which was minimized from `crux-llvm`'s [`strlen_test2.c`](https://github.com/GaloisInc/crucible/blob/2b6062ce47151bb44f794694358d6f0d0f120774/crux-llvm/test-data/golden/golden/strlen_test2.c):

```c
// test.c
#include <stdint.h>
#include <stdio.h>

__attribute__((noinline)) void f(const char *string, uint32_t length) {
    uint32_t i;
    for (i = 0; i < length && i < 200; i++) {
        printf("%u\n", string[i]);
    }
}

int main(void) {
    char str[6] = "Hello";
    f(str, 6);
    return 0;
}
```

To trigger the bug, you must compile the bitcode with optimizations enabled using LLVM 18 or later:

```
$ clang -g -emit-llvm test.c -c -O
```

Parsing this yields the following error:

```
λ> parseBitCodeFromFileWithWarnings "test.bc" >>= putStrLn . either formatError show
Invalid record size: 4
Expected one of: [3]

Are you sure you're using a supported compiler?
Check here: https://github.com/GaloisInc/llvm-pretty-bc-parser

from:
	FUNC_CODE_INST_CAST
	@f
	FUNCTION_BLOCK
	FUNCTION_BLOCK_ID
	value symbol table
	MODULE_BLOCK
	Bitstream
```

Fixing this will require updates to `llvm-pretty` (to store the `nneg` flag), `llvm-pretty-bc-parser` (to parse `nneg` flags), and `crucible-llvm` (to ensure that the semantics of `nneg` are respected).

RyanGlScott:
Fixes #306.

### DIFF
--- a/src/Data/LLVM/BitCode/IR/Function.hs
+++ b/src/Data/LLVM/BitCode/IR/Function.hs
@@ -431,6 +431,9 @@ parseFunctionBlockEntry _ t d (fromEntry -> Just r) =
     -- - If the instruction is trunc, the extra field designates the value of
     --   the nuw and nsw flags.
     --
+    -- - If the instruction is fptrunc or fpext, the extra field designates the
+    --   value of the fast-math flags. TODO(#61): We currently ignore these.
+    --
     -- The constructor returned from castOp will use that value when
     -- constructing the cast-related operation.
     let mbWord = numeric =<< fieldAt (ix+2) r


src/Data/LLVM/BitCode/IR/Constants.hs
```diff
@@ -152,31 +152,56 @@ unopCE = unopGeneric uaop
...
+  nuw x = testBit x 0
+  nsw x = testBit x 1
+
+  -- operations that accept the nuw and nsw flags
+  wrapFlags c = return $ \mb tv t ->
+    case mb of
+      Nothing -> op (c False False) tv t
+      Just w -> op (c (nuw w) (nsw w)) tv t
+
+  exact x = testBit x 0
```
kquick: Should this be renamed from `exact` to `nneg` to match lines 163 and 164 above?

src/Data/LLVM/BitCode/IR/Constants.hs
```diff
@@ -152,31 +152,56 @@ unopCE = unopGeneric uaop
...
-castOp = castOpGeneric (return . Conv)
+castOp :: Match Field (Maybe Int -> Typed PValue -> Type -> PInstr)
+castOp = castOpGeneric Conv
 
+-- NB: Unlike the type of 'castOp', this does not offer a @Maybe Int@ argument
+-- in its callback. This is because at present, LLVM constant expressions do not
+-- accept flags like their expression counterparts do.
 castOpCE :: Match Field (Typed PValue -> Type -> PValue)
-castOpCE = castOpGeneric op
+castOpCE i = (\p -> p Nothing) <$> castOpGeneric op i
```
kquick: Neat trick.  IIUC, this could probably also be:
 `castOpCE i = ($ Nothing) <$> castOpGeneric op i`
 or perhaps even
 `castOpCE = fmap ($ Nothing) . castOpGeneric op`
 if one were golfing.

kquick (approved):

src/Data/LLVM/BitCode/IR/Constants.hs
kquick: The only teeny tiny difference is that `($ x)` can only do one thing, whereas `\p ->` needs a little more scrutiny to see what it's doing.  That said, `($ x)` is probably not well known syntax either, and I was just having fun exploring this new (to me) functor, so no change necessary.

PR #322, opened 2025-11-14
Add some tests validating the inrange and inbounds flags.

kquick:


RyanGlScott (approved):

RyanGlScott:
For what it's worth, I think these are also broken on `master`, as seen in [this scheduled `master` CI job](https://github.com/GaloisInc/llvm-pretty-bc-parser/actions/runs/19371875186).

langston-barrett:
Also FWIW if you're not sure what's causing the problem, there is an official LLVM testcase reducer that could help: https://llvm.org/docs/CommandGuide/llvm-reduce.html

PR #323, opened 2025-11-15
Fix disasm-test range handling and re-introduce T266.

kquick:
This fixes the disasm-test's handling of `pre-llvmX` and `post-llvmX` to work as expected.

Re-introduces `T266` which was temporarily dropped in https://github.com/GaloisInc/llvm-pretty-bc-parser/commit/9619e615d45d4cb838726ba743b1e9b14da61269 due to the issues that are fixed here.

### DIFF
--- a/disasm-test/Main.hs
+++ b/disasm-test/Main.hs
@@ -451,10 +451,6 @@ rangeMatch llvmver cb swts = do
                                -- filtered, but make this total.
                                (e:_) -> lookup "llvm-range" $ TS.expParamsMatch e
                                [] -> Nothing
-            invert = \case
-              LT -> GT
-              EQ -> EQ
-              GT -> LT
             better s1 s2 =
               -- if both pre-llvm, lower is better (sort is lowest-to-highest),
               -- and if both post-llvm, higher is better; if both match, either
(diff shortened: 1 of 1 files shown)


disasm-test/Main.hs
```diff
@@ -416,26 +417,72 @@ assemblyCube rootPath llvmver = TS.mkCUBE
...
+             (al, bl) -> compare al bl
+  let selectFromGroup :: [TS.Sweets] -> [TS.Sweets]
+      selectFromGroup = take 1 -- only the first element of each group
+                        . sortBy tightestLLVMMatch -- preferred ordering
+                        -- remove any group member with no targets
+                        . filter (not . null . TS.expected)
+  let ts3 = concat
+            $ fmap selectFromGroup
+            $ groupBy ((==) `on` TS.rootBaseName) ts2 -- group by base name
+  return ts3
```
RyanGlScott: Perhaps:


```suggestion
  concat
    $ fmap selectFromGroup
    $ groupBy ((==) `on` TS.rootBaseName) ts2 -- group by base name
```

RyanGlScott (approved):

PR #324, opened 2025-11-15
Add documentation for nix CI updates of GHC and LLVM versions.

### LINKED ISSUE #276: Document how to update `flake.lock` file when bumping `llvm-pretty` submodule
The `llvm-pretty-bc-parser` CI uses a Nix flake, and the associated [`flake.lock` file](https://github.com/GaloisInc/llvm-pretty-bc-parser/actions/runs/10599192545/job/29373612312) needs to be updated every time the `llvm-pretty` submodule commit is updated. If you forget to do this, the CI will likely fail with mysterious errors due to the Nix-based CI checkout out a different `llvm-pretty` commit than what the submodule targets. (See [here](https://github.com/GaloisInc/llvm-pretty-bc-parser/actions/runs/10599192545/job/29373612312) for one example of this.) This has led to quite a bit of head-scratching (between myself, @mccleeary-galois, and others) when trying to debug CI failures.

We should clearly document what Nix command you need to run in order to update the `flake.lock` file in tandem with a submodule bump. (I vaguely recall that it's something to the effect of `nix flake update`, but the specifics escape me.)

kquick:
Fixes #276.

### DIFF
--- a/doc/developing.md
+++ b/doc/developing.md
@@ -98,19 +98,21 @@ When updating the supported GHC versions, remember to update:
 - [The Cabal file](../llvm-pretty-bc-parser.cabal)'s `Tested-with` field
 - [The Nix flake](../flake.nix)
 
-  Run `nix flake update` (installing [`nix`](https://nixos.org) first if
-  necessary) which will update the `flake.lock` file in the top level directory.
+  Run `nix flake update [input-pkg ...]` (installing [`nix`](https://nixos.org)
+  first if necessary) which will update the `flake.lock` file in the top level
+  directory.  If no `input-pkg` is specified, *all* inputs are updated.
   Optionally run `nix build ./#TESTS` to run the nix-based tests locally.  Then
   commit the updated `flake.lock` file.
 
   Newer GHC versions are automatically available to the nix build, but may
-  require updating the flake lockfile as described above to become visible.  To
-  add new GHC versions to the `nix` builds, see the `ghc-version` list in the
-  `nix-ci.yml` file.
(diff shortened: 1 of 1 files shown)


doc/developing.md
```diff
@@ -97,4 +97,23 @@ When updating the supported GHC versions, remember to update:
 - [The README](../README.md)
 - [The Cabal file](../llvm-pretty-bc-parser.cabal)'s `Tested-with` field
 - [The Nix flake](../flake.nix)
+
+  Run `nix flake update` (installing [`nix`](https://nixos.org) first if
+  necessary) which will update the `flake.lock` file in the top level directory.
+  Optionally run `nix build ./#TESTS` to run the nix-based tests locally.  Then
+  commit the updated `flake.lock` file.
```
RyanGlScott: My understanding is that `nix flake update` will update _everything_ in the lock file to the latest version, is that correct? What would be the workflow if you only wanted to update a single dependency (e.g., the `llvm-pretty` submodule)?

doc/developing.md
```diff
@@ -97,4 +97,23 @@ When updating the supported GHC versions, remember to update:
...
+
+  Newer GHC versions are automatically available to the nix build, but may
+  require updating the flake lockfile as described above to become visible.  To
+  add new GHC versions to the `nix` builds, see the `ghc-version` list in the
+  `nix-ci.yml` file.
```
langston-barrett: Minor point, but to me it seems like this last sentence (starting with "To") is what developers should try first, and the rest is details on what to do if that doesn't work. Perhaps this sentence could be the first part of this section?

PR #328, opened 2025-11-18
Use tree-diff to report direct AST differences in disasm-test.

kquick:
This will allow the direct comparison of _normalized_ AST structures instead of just a file-based diff of the Show of those structures (which is what was done up until this change).

Doing a direct comparison allows for normalization to address elements that are not particularly important or are equivalent, but have different representations. For example, LLVM 21 changes the Size and Offset fields in DIBasicType, DICompositeType, DIDerivedType, and several others to be either a direct integer or a metadata reference; a metadata reference with a size of 8 should be equal to a direct integer of 8, and the normalization can address this.

### DIFF
--- a/llvm-pretty-bc-parser.cabal
+++ b/llvm-pretty-bc-parser.cabal
@@ -127,6 +127,7 @@ test-suite unit-test
 Test-suite disasm-test
   type:                exitcode-stdio-1.0
   Main-is:             Main.hs
+  other-modules:       Instances
   Default-language:    Haskell2010
   hs-source-dirs:      disasm-test
   Ghc-options:         -Wall


disasm-test/Main.hs
```diff
@@ -528,6 +534,104 @@ runAssemblyTest llvmVersion knownBugs sweet expct
                        Nothing -> error "Failed processLL"
 
 
+
+-- | Compare two ASTs to see if they are the same.  Fundamentally is just done
+--  via (==) on the normalized ASTs, but this first uses the tree-diff package to
+--  generate an nicer output to allow focus on the actual diffs rather than
```
RyanGlScott:
```suggestion
--  generate nicer output to allow focusing on the actual diffs rather than
```

disasm-test/Main.hs
```diff
@@ -528,6 +534,104 @@ runAssemblyTest llvmVersion knownBugs sweet expct
...
+--  leaving it to the user to analyze the large blobs to find out where.
+cmpASTs :: AST.Module -> AST.Module -> TestM ()
+cmpASTs ast1 ast2 = do
+  let d = ediff ast1 ast2
+      msg = "Differences (marked with + and - line prefixes:\n"
+            <> show (prettyEditExprCompact d)
+  liftIO $ assertBool msg $ ast1 == ast2
+
+deriving instance ToExpr (AST.Type' AST.Ident)
```
RyanGlScott: It might be worth moving these orphan instances to their own module (akin to `unit-test/Tests/ExpressionInstances.hs`) to reduce the time it takes to rebuild `disasm-test/Main.hs` on its own.

PR #329, opened 2025-11-18
Use common parsing of DebugLoc.

kquick:
Reduces duplications to avoid errors, and this will be useful for LLVM 21's additions to DebugLoc.

Note that the actual parsing routine needs to be fairly polymorphic: one call uses `Int` as the `r` parameter and the other uses `BlockLabel`, and the helpers for those are slightly different, thus they are passed as arguments.

### DIFF
--- a/src/Data/LLVM/BitCode/IR/Metadata.hs
+++ b/src/Data/LLVM/BitCode/IR/Metadata.hs
@@ -1213,10 +1213,11 @@ parseMetadataOldNode fnLocal vt mt r pm = do
 
     _ -> fail "Malformed metadata node"
 
-parseDebugLoc :: Num a => Bits a => Num b => Bits b
+parseDebugLoc :: Num scope => Bits scope => Num ia => Bits ia
               => Int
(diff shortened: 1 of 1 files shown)


src/Data/LLVM/BitCode/IR/Metadata.hs
```diff
@@ -1251,6 +1241,26 @@ parseMetadataOldNode fnLocal vt mt r pm = do
 
     _ -> fail "Malformed metadata node"
 
+parseDebugLoc :: Num a => Bits a => Num b => Bits b
+              => Int
+              -> (a -> Parse (ValMd' r))
+              -> (b -> Parse (Maybe (ValMd' r))) -> Record
+              -> Parse (DebugLoc' r)
```
RyanGlScott: Style nit pick: put each argument on its own line:

```hs
            -> (b -> Parse (Maybe (ValMd' r)))
            -> Record
            -> Parse (DebugLoc' r)
```

RyanGlScott (approved):

PR #345, opened 2026-01-30
Add support for more floating-point types

sauclovian-g:
Pulls in https://github.com/GaloisInc/llvm-pretty/pull/186 and adds code to read all the missing float types.

Pretty much untested, alas.

Provisional until the upstream PR merges.

### DIFF
--- a/disasm-test/tests/constantfloat.ll
+++ b/disasm-test/tests/constantfloat.ll
@@ -14,15 +14,14 @@
 ; prints a diff of what you got wrong... with no visible differences,
 ; which is highly confusing.
 ;
-; Also disable the inf case until someone fixes llvm-pretty #187. We
-; print it as 0x7f800000; apparently the proper behavior for printing
-; single-precision as hex is to promote to double and print that.
-; (This is also why the hex values below are doubles.)
+; Note: apparently the proper behavior for printing single-precision
+; as hex is to promote to double and print that. This is also why the
+; hex values below are doubles.
 @f0 = dso_local global float 0.0, align 4
 @f1 = dso_local global float 1.0, align 4
 @f2 = dso_local global float 2.0, align 4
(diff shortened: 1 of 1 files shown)


sauclovian-g:
ah, the nix stuff isn't going to work without updating the llvm-pretty reference. I'm going to hold off on that until it's final, I think, so I don't have to do it twice.

sauclovian-g:
I figured out the tests :-)

Note that we can't test NaNs in either single or double precision because we store them as `Float` and `Double` respectively in the internal representation. Because NaNs are not equal to themselves, the correct AST is also not equal to itself, so the test fails. And then you get an empty diff out, which is quite confusing...

Also we can't test single-precision floats that need to be printed as hex (which, since NaNs don't work, amounts to +/- Inf) because we're doing it wrong. See https://github.com/GaloisInc/llvm-pretty/issues/187.

disasm-test/tests/constantfloat.ll
```diff
@@ -0,0 +1,58 @@
...
+; Do not attempt to test NaNs; it breaks the test infrastructure.
+; Because we load these into the Haskell float type, a NaN is not
+; equal to itself, and then the test both (a) fails and then (b)
+; prints a diff of what you got wrong... with no visible differences,
+; which is highly confusing.
+;
+; Also disable the inf case until someone fixes llvm-pretty #187. We
+; print it as 0x7f800000; apparently the proper behavior for printing
+; single-precision as hex is to promote to double and print that.
+; (This is also why the hex values below are doubles.)
```
RyanGlScott: Perhaps we should wait to rebase this once https://github.com/GaloisInc/llvm-pretty/pull/188 has landed in `llvm-pretty` so that we can uncomment these cases.

RyanGlScott (approved):

Issue #340, opened 2026-01-19
Support `DIBasicType`'s `dataSize` field (introduced in LLVM 22)


RyanGlScott:
LLVM 22 adds a `dataSize` field to `DIBasicType`. See https://github.com/llvm/llvm-project/commit/aa5fe56db4777dc1dbd8e114090711068e76c770 for the commit which introduces this. Currently, `llvm-pretty-bc-parser` emits a warning because of its lack of `dataSize` support. To see this, compile the following C file:

```c
// test.c
int main(void) {
  return 0;
}
```

Using `clang-22 test.c -g -emit-llvm -c`, then load the resulting bitcode:

```
λ> :m + Text.LLVM.PP Data.LLVM.BitCode
λ> parseBitCodeFromFileWithWarnings "test.bc" >>= either (putStrLn . formatError) (print . ppParseWarnings . snd)
Encountered the following warnings during parsing:
    Invalid record size: 23
    Expected size between 14 and 22
    from:
    	METADATA_COMPILE_UNIT
    	METADATA_BLOCK
    	METADATA_BLOCK_ID
    	value symbol table
    	MODULE_BLOCK
    	Bitstream

    Invalid record size: 9
    Expected size between 6 and 8
    from:
    	METADATA_BASIC_TYPE
    	METADATA_BLOCK
    	@main
    	FUNCTION_BLOCK
    	FUNCTION_BLOCK_ID
    	value symbol table
    	MODULE_BLOCK
    	Bitstream

Are you sure you're using a supported version of LLVM/Clang?
Check here: https://github.com/GaloisInc/llvm-pretty-bc-parser
```

(Note that the `METADATA_COMPILE_UNIT` warnings are unrelated and are tracked separately in https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/339.)

Issue #339, opened 2026-01-19
Support `DICompileUnit`'s `sourceLanguageVersion` field (introduced in LLVM 22)


RyanGlScott:
LLVM 22 introduces a `sourceLanguageVersion` field to `DICompileUnit`. See https://github.com/llvm/llvm-project/commit/cf1cdde24e76a4ff5bbda2128800ba3ca4132b6d for the LLVM commit which introduces this. Currently, `llvm-pretty-bc-parser` emits a warning because of its lack of `sourceLanguageVersion` support. To see this, compile an empty C file using `clang-22 -g -emit-llvm -c` and load the resulting bitcode file:

```
λ> :m + Text.LLVM.PP Data.LLVM.BitCode
λ> parseBitCodeFromFileWithWarnings "test.bc" >>= either (putStrLn . formatError) (print . ppParseWarnings . snd)
Encountered the following warnings during parsing:
    Invalid record size: 23
    Expected size between 14 and 22
    from:
    	METADATA_COMPILE_UNIT
    	METADATA_BLOCK
    	METADATA_BLOCK_ID
    	value symbol table
    	MODULE_BLOCK
    	Bitstream

Are you sure you're using a supported version of LLVM/Clang?
Check here: https://github.com/GaloisInc/llvm-pretty-bc-parser
```

PR #347, opened 2026-02-25
Support parsing metadata fields added in LLVM 22

RyanGlScott:
This adds support for parsing the `sourceLanguageVersion` and `dataSize` fields, which were introduced in LLVM 22. As such, this fixes https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/339 and fixes https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/340, respectively.

This bumps the `llvm-pretty` submodule to bring in the corresponding changes from https://github.com/GaloisInc/llvm-pretty/pull/189.

### DIFF
--- a/.github/workflows/nix-ci.yml
+++ b/.github/workflows/nix-ci.yml
@@ -7,6 +7,10 @@ on:
   pull_request:
   workflow_dispatch:
 
+defaults:
+  run:
+    shell: bash
+
 jobs:
   build:
     # This job builds the main package, via a number of GHC versions.  This is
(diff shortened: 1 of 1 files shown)


.github/workflows/nix-ci.yml
```diff
@@ -84,6 +84,23 @@ jobs:
...
+      - uses: cachix/install-nix-action@v20
+        with:
+          nix_path: nixpkgs=channel:nixos-23.05
+          github_access_token: ${{ secrets.GITHUB_TOKEN }}
+      - uses: cachix/cachix-action@v15
+        with:
+          name: galois
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+      - name: test
+        shell: bash
```
langston-barrett: It would be good to do this at the top level in a `defaults` stanza: https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/set-default-values-for-jobs

PR #352, opened 2026-04-07
Changes to support the UnnamedMdIdx newtype wrapper in llvm-pretty

kquick:
This implements changes for the newtype wrapper around the unnamed metadata index.   This adds clarity and individuality to that type, which requires some conversions here from general indices that could apply to multiple types to a refinement of that index into a metadata-specific type.

### DIFF
--- a/src/Data/LLVM/BitCode/IR/Metadata.hs
+++ b/src/Data/LLVM/BitCode/IR/Metadata.hs
@@ -193,13 +193,13 @@ mdStringOrNull :: HasCallStack
                -> UnnamedMdIdx
                -> Maybe String
 mdStringOrNull cxt partialMeta ix =
-  let pmEntryStr = \case
-        ValMdString str -> str
-        _ -> let explanation = "Non-string metadata when string was expected"
-             in throw (BadTypeRef callStack cxt explanation $ unnamedMdIdx ix)
-  in do aix <- nonNullUnnamedMdIdx ix
-        Map.lookup aix (pmStrings partialMeta)
-          <|> (pmEntryStr <$> mdForwardRefOrNull cxt (pmEntries partialMeta) ix)
+  do aix <- nonNullUnnamedMdIdx ix
+     Map.lookup aix (pmStrings partialMeta)
+       <|> case mdForwardRef cxt (pmEntries partialMeta) aix of
+             ValMdString str -> Just str
+             _ -> let explanation = "Non-string metadata when string was expected"
+                  in throw (BadTypeRef callStack cxt explanation $ unnamedMdIdx ix)
+
 
 mdStringOrEmpty :: HasCallStack
                 => [String]
--- a/src/Data/LLVM/BitCode/IR/Metadata.hs
+++ b/src/Data/LLVM/BitCode/IR/Metadata.hs
@@ -40,6 +40,7 @@ import           Control.Exception (throw)
 import           Control.Monad (foldM, guard, mplus, unless, when)
 import           Data.Bits ( Bits, shiftR, testBit, shiftL, (.&.), (.|.), bit
                            , complement )
+import           Data.Bool ( bool )
 import qualified Data.ByteString as S
 import qualified Data.ByteString.Char8 as Char8 (unpack)
 import           Data.Data (Data)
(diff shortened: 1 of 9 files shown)


src/Data/LLVM/BitCode/IR/Metadata.hs
```diff
@@ -148,56 +156,57 @@ addOldNode fnLocal vals mt = nameNode fnLocal False ix mt'
...
     let explanation = "Illegal forward reference into function-local metadata."
-    in throw (BadValueRef callStack cxt explanation r)
+    in throw (BadValueRef callStack cxt explanation $ unnamedMdIdx r)
 
-mdForwardRefOrNull :: [String] -> MetadataTable -> Int -> Maybe PValMd
-mdForwardRefOrNull cxt mt ix | ix > 0 = Just (mdForwardRef cxt mt (ix - 1))
-                             | otherwise = Nothing
+mdForwardRefOrNull :: [String] -> MetadataTable -> UnnamedMdIdx -> Maybe PValMd
+mdForwardRefOrNull cxt mt ix
+  | unnamedMdIdx ix > 0 = Just (mdForwardRef cxt mt (nonNullUnnamedMdIdx ix))
```
RyanGlScott: The `UnnamedMdIdx` abstraction feels particularly leaky here. In order to tell whether it's null or not, you have to directly compare the underlying `Int` value against `0`. Rather than doing this, I wonder if we should instead define `nonNullUnnamedMdIdx` to have the type `UnnamedMdIdx -> Maybe UnnamedMdIdx`, where it returns `Nothing` if the supplied `UnnamedMdIdx` is not present (i.e., if it is `0`) and it returns `Just` otherwise. Then we could define `mdForwardRefOrNull` as:

```hs
mdForwardRefOrNull :: [String] -> MetadataTable -> UnnamedMdIdx -> Maybe PValMd
mdForwardRefOrNull cxt mt ix = mdForwardRef cxt mt <$> nonNullUnnamedMdIdx ix
```

src/Data/LLVM/BitCode/IR/Metadata.hs
```diff
@@ -902,15 +912,15 @@ parseMetadataEntry vt mt pm (fromEntry -> Just r) =
     24 -> label "METADATA_NAMESPACE" $ do
       assertRecordSizeIn r [3, 5]
       let isNew = length (recordFields r) == 3
-      let nameIdx = if isNew then 2 else 3
+      let nameIdx = bool 3 2 isNew
```
RyanGlScott, line 913: Style nitpick: cosmetic changes like refactoring `if isNew then 2 else 3` to `bool 3 2 isNew` don't have anything to do with `UnnamedMdIdx`, and for ease of reviewing, it would be helpful to split these changes out into a separate commit.

src/Data/LLVM/BitCode/IR/Metadata.hs
```diff
@@ -195,13 +193,13 @@ mdStringOrNull :: HasCallStack
...
-      Just _                 ->
-        let explanation = "Non-string metadata when string was expected"
-        in throw (BadTypeRef callStack cxt explanation $ unnamedMdIdx ix)
+  let pmEntryStr = \case
+        ValMdString str -> str
+        _ -> let explanation = "Non-string metadata when string was expected"
+             in throw (BadTypeRef callStack cxt explanation $ unnamedMdIdx ix)
+  in do aix <- nonNullUnnamedMdIdx ix
+        Map.lookup aix (pmStrings partialMeta)
+          <|> (pmEntryStr <$> mdForwardRefOrNull cxt (pmEntries partialMeta) ix)
```
RyanGlScott: I think there's some duplicate work here: we call `nonNullUnnamedMdIdx ix` first, but later call `mdForwardRefOrNull cxt (pmEntries partialMeta) ix`, which in turns calls `nonNullUnnamedMdIdx ix` again. At this point, you know that `nonNullUnnamedMdIdx ix` returned `Just aix`, so it should be cheaper to do something like:

```hs
  in do aix <- nonNullUnnamedMdIdx ix
        Map.lookup aix (pmStrings partialMeta)
          <|> case mdForwardRef cxt (pmEntries partialMeta) aix of
                ValMdString str -> str
                _ -> let explanation = "Non-string metadata when string was expected"
                     in throw (BadTypeRef callStack cxt explanation $ unnamedMdIdx ix)
```

flake.nix
```diff
@@ -20,7 +20,7 @@
       inputs.nixpkgs.follows = "nixpkgs";
     };
     llvm-pretty-src = {
-      url = "github:GaloisInc/llvm-pretty";
+      url = "github:GaloisInc/llvm-pretty/umdidx_newtype";
```
RyanGlScott: Please make sure to switch back to the main branch of `llvm-pretty` before landing this.

RyanGlScott (approved):

PR #353, opened 2026-04-22
Add support for DISubrangeType (LLVM 21)

kquick:


disasm-test/tests/disubrangetype.ll
```diff
@@ -0,0 +1,25 @@
+;; This test checks generation of DISubrangeType.
+;; Copy of llvm-project llvm/test/Bitcode/subrange_type.ll
+
+; ModuleID = 'subrange_type.ll'
+source_filename = "/dir/subrange_type.adb"
```
RyanGlScott, line 5: I assume that this `.ll` file was generated from a `subrange_type.adb` Ada file? If so, I think it would be worth checking in the Ada file such that we could regenerate this `.ll` file in the future if need be.

disasm-test/tests/disubrangetype.pre-llvm22.ll
RyanGlScott, line 1: Per https://github.com/GaloisInc/llvm-pretty/pull/196#discussion_r3124625773, I think this should be `pre-llvm21`, not `pre-llvm22`.

flake.nix
```diff
@@ -20,7 +20,7 @@
       inputs.nixpkgs.follows = "nixpkgs";
     };
     llvm-pretty-src = {
-      url = "github:GaloisInc/llvm-pretty";
+      url = "github:GaloisInc/llvm-pretty/de6acc11";
```
RyanGlScott: Make sure to revert this change before landing.

disasm-test/tests/disubrangetype.ll
kquick: It probably was, but that was not included in the upstream commit in the llvm-project: just the .ll file.  I can write an Ada file (probably generating a different example), but we will need to have an active AdaCore license to be able to run the Ada->LLVM compiler to regenerate the `.ll` file, which is not something we will likely have long-term.

RyanGlScott: Ah, OK! In that case, perhaps just linking to where you obtained the `.ll` file upstream would suffice.

src/Data/LLVM/BitCode/IR/Metadata.hs
```diff
@@ -421,6 +421,17 @@ parseMetadataEntry vt mt pm (fromEntry -> Just r) =
                  mdForwardRefOrNull ctx mt <$> parseMdIdx r n
       ronl n = if length (recordFields r) <= n then pure Nothing else ron n
 
+      asSignedVal = fmap $ \case
```
RyanGlScott, line 426: I think the name of this function ought to mention "64" somewhere in the name, as this really only works if the underlying integer encodes a 64-bit integer. It would also be worth adding some comments explaining what this is doing (namely, it converts from a two's complement representation to an explicitly signed one).

src/Data/LLVM/BitCode/IR/Metadata.hs
```diff
@@ -421,6 +421,17 @@ parseMetadataEntry vt mt pm (fromEntry -> Just r) =
...
+        v@(ValMdValue tv)
+          | PrimType (Integer s) <- typedType tv
+          , ValInteger i <- typedValue tv
+            -> let x = fromInteger i :: Word64
+                   xNeg = (toInteger $ (complement x + 1)) * (-1)
```
RyanGlScott: Style nit pick: fewer parentheses, fewer infix operators:


```suggestion
                   xNeg = toInteger (complement x + 1) * (-1)
```

If you want to keep the outer parentheses for aesthetic reasons, you could also write:


```suggestion
                   xNeg = -(toInteger (complement x + 1))
```

RyanGlScott (approved):
Make sure to mention these changes in the changelog.

src/Data/LLVM/BitCode/IR/Metadata.hs
kquick: I actually made it more general for other sizes and documented as such.

PR #354, opened 2026-04-23
Adds parsing for DIFixedPointType (metadata ID 49) from LLVM 21

kquick:


src/Data/LLVM/BitCode/IR/Metadata.hs
```diff
@@ -1243,6 +1243,41 @@ parseMetadataEntry vt mt pm (fromEntry -> Just r) =
...
+      difptTag <- parseField r 1 numeric
+      difptName <- mdStringOrNull ctx pm <$> parseMdIdx r 2
+      difptSize <- ron 3
+      difptAlign <- parseField r 4 numeric
+      difptEncoding <- parseField r 5 numeric
+      difptFlags <- parseField r 6 numeric
+      kind <- parseField r 7 numeric
+      difptKind <-
+        case (kind :: Integer) of
+          0 -> FixedPointBinary <$> parseField r 8 numeric
```
RyanGlScott, line 1260: Should this (and the fields below) be parsed as signed integers? [This code](https://github.com/llvm/llvm-project/blob/68947342b75cc71f3ac9041d11db086d8d074336/llvm/lib/AsmParser/LLParser.cpp#L5557-L5559) in LLVM leads me to believe that they should be.

unit-test/Tests/ExpressionInstances.hs
```diff
@@ -35,6 +35,7 @@ instance Arbitrary lab => Arbitrary (DISubrangeType' lab)             where arbi
 instance Arbitrary lab => Arbitrary (DICompileUnit' lab)              where arbitrary = genericArbitrary uniform
 instance Arbitrary lab => Arbitrary (DICompositeType' lab)            where arbitrary = genericArbitrary uniform
 instance Arbitrary lab => Arbitrary (DIDerivedType' lab)              where arbitrary = genericArbitrary uniform
+instance Arbitrary lab => Arbitrary (DIFixedPointType' lab)           where arbitrary = genericArbitrary uniform
```
RyanGlScott, line 38: Make sure to derive an instance for `DIFixedPointKind'` as well.

RyanGlScott (approved):

src/Data/LLVM/BitCode/IR/Metadata.hs
kquick: They are, as confirmed by the added test which does use negative values.

RyanGlScott: Oh, my mistake, I had assumed that all integers needed to be parsed using `asSignedVal`. Upon further inspection, that really only applied to `ValMd`s, and not to integers parsed using `parseField ... numeric`.

PR #356, opened 2026-04-23
Generates a useful error when no LLVM tools are present.

kquick:
Previously, running the `disasm-test` when there were no LLVM tools installed (`llvm-as`, `llvm-dis`, and `clang`), the error output was:

```
disasm-test: Enum.pred{Word}: tried to take `pred' of minBound
```

which is mysterious and unhelpful.  With this change, the error is

```
disasm-test: Unexpected version mismatch between clang, llvm-as and llvm-dis
* llvm-as  version: llvm-as [missing]
* llvm-dis version: llvm-dis [missing]
* clang    version: clang [missing]
```

followed by a callstack/`error` exit report, which is much more helpful and diagnosable by the user.

### DIFF
--- a/disasm-test/Main.hs
+++ b/disasm-test/Main.hs
@@ -24,7 +24,7 @@ import qualified Data.ByteString.Lazy as L
 import           Data.Char (ord,isLetter,isSpace,chr)
 import           Data.Function ( on )
 import           Data.Generics (everywhere, mkT) -- SYB
-import           Data.List ( find, isInfixOf, isPrefixOf, isSuffixOf, nub, sort , groupBy, sortBy )
+import           Data.List ( find, isPrefixOf, isSuffixOf, nub, sort , groupBy, sortBy )
 import           Data.Map ( (!), (!?) )
 import qualified Data.Map as Map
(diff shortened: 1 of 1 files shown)


disasm-test/Main.hs
```diff
@@ -346,6 +346,7 @@ main =  do
   clangVC <- getClangVersion clang'
   unless (and [ vcVersioning llvmAsVC == vcVersioning llvmDisVC
               , vcVersioning llvmAsVC == vcVersioning clangVC
+              , not ("missing" `isInfixOf` showVC llvmAsVC)
```
RyanGlScott: I take it that the word "missing" arises from the definition of `readProcessVersion` ([here](https://github.com/GaloisInc/llvm-pretty-bc-parser/blob/f7cccd80581d4741393ab93fe82d71a9d49d25b0/disasm-test/Main.hs#L315))? If so, wouldn't a more direct check be something like `vcVersioning llvmAsVC == Left "[missing]"`? (We could also factor out the `"[missing]"` string into its own definition to ensure that the two stay in sync.)

RyanGlScott (approved):

PR #357, opened 2026-04-24
LLVM Metadata parse failures are warnings only

### LINKED ISSUE #355: Downgrade unsupported metadata record codes from errors to warnings
We have made some strides towards making such that `llvm-pretty-bc-parser` doesn't emit a fatal error if you encounter a form of metadata that isn't currently supported (see https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/248). Unfortunately, this only covers the case where a known metadata type is parsed, but with a different number of records than expected. If you encounter an entirely new metadata record code, on the other hand, then `llvm-pretty-bc-parser` will still treat that as a hard error.

As an example, consider what happens if you parse a bitcode file produced from the `disasm-test/tests/difixedpointtype.ll` test case added in #354, but using an `llvm-pretty-bc-parser` commit that predates #354 (e.g., 6f49d2c295a9f5676e798a4f189da0a274691134). This commit lacks support for parsing metadata record code 49 (`METADATA_FIXED_POINT_TYPE`), and this causes `llvm-pretty-bc-parser` to fail with the following error:

```
λ> :m + Text.LLVM.PP Data.LLVM.BitCode
λ> parseBitCodeFromFileWithWarnings "test.bc" >>= either (putStrLn . formatError) (print . ppLLVM llvmVlatest ppModule . fst)
unknown record code: 49
from:
	METADATA_BLOCK
	METADATA_BLOCK_ID
	value symbol table
	MODULE_BLOCK
	Bitstream
```

Like with #248, we should downgrade this from an error to a warning.

kquick:
Fixes #355.

### DIFF
--- a/src/Data/LLVM/BitCode/IR/Metadata.hs
+++ b/src/Data/LLVM/BitCode/IR/Metadata.hs
@@ -455,6 +455,15 @@ parseMetadataEntry vt mt pm (fromEntry -> Just r) =
             (Just . ValMdValue . Typed (PrimType (Integer 64)) . ValInteger)
             (parseField r n numeric)
 
+      mdNotImplemented :: String -> Parse PartialMetadata
+      mdNotImplemented name = do
+        addParseWarning $ UnsupportedMetadataRecordCode (recordCode r) name
+        -- Some sort of placeholder entry must be added to the metadata table,
+        -- because it may be cross-reference by other metadata; here we use an
+        -- empty expression as that placeholder.
+        return $! updateMetadataTable
+          (addDebugInfo False (DebugInfoExpression $ DIExpression mempty)) pm
+
   -- Note: the parsing cases below use a Monadic coding style, as opposed to an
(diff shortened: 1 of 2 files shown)


src/Data/LLVM/BitCode/IR/Metadata.hs
```diff
@@ -455,6 +455,17 @@ parseMetadataEntry vt mt pm (fromEntry -> Just r) =
             (Just . ValMdValue . Typed (PrimType (Integer 64)) . ValInteger)
             (parseField r n numeric)
 
+      -- Used when some sort of placeholder entry must be added to the metadata
+      -- table, because it may be cross-reference by other metadata; here we use
```
RyanGlScott:
```suggestion
      -- table, because it may be cross-referenced by other metadata; here we use
```

RyanGlScott (approved):

PR #358, opened 2026-04-25
Parse constant integer values as signed.

kquick:
Prior to this, the constant values were parsed and stored in their raw, two's-complement unsigned value.  This was visible as seen in the following:

```
$ cabal run disasm-test -- -p p0 --details true | grep -A4'__const.main.msg .* constant'
...
--
-@__const.main.msg = private constant { i16, [6 x i8], ptr, i8, [7 x i8] } { i16 0, [6 x i8] zeroinitializer, ptr @signal, i8 -56, [7 x i8] zeroinitializer }, align 8
-
-; Function Attrs: nocallback nofree nounwind willreturn memory(argmem: readwrite)
-declare void @llvm.memcpy.p0.p0.i64(ptr noalias writeonly captures(none), ptr noalias readonly captures(none), i64, i1 immarg) #0
-
--
+@__const.main.msg = private default constant { i16, [6 x i8], ptr,
+                                               i8, [7 x i8] } { i16 0, [6 x i8] zeroinitializer,
+                                                                ptr @signal,
+                                                                i8 18446744073709551560,
+                                                                [7 x i8] zeroinitializer }, align 8

```

Note that the first initialization shown (the output from LLVM's `llvm-dis`) specifies the third value as `i8 -56`, whereas the second initialization shown (the output from this package's `llvm-disasm`) specifies that same value as `i8 18446744073709551560`, which is silly and brittle.

With this change, the parsing properly converts the two's-complement unsigned integer value into a signed value and the `llvm-disasm` output of these values now matches that of `llvm-dis`.

### DIFF
--- a/src/Data/LLVM/BitCode/Parse.hs
+++ b/src/Data/LLVM/BitCode/Parse.hs
@@ -2,11 +2,12 @@
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
-{-# LANGUAGE TypeSynonymInstances #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE RecursiveDo #-}
 {-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE RecursiveDo #-}
+{-# LANGUAGE TypeSynonymInstances #-}
 
 module Data.LLVM.BitCode.Parse where
 
@@ -24,11 +25,12 @@ import           Control.Monad.Except (MonadError(..), Except, runExcept)
 import           Control.Monad.Reader (MonadReader(..), ReaderT(..), asks)
 import           Control.Monad.State.Strict (MonadState(..), StateT(..)
                                             , gets, modify)
+import           Data.Bits ( testBit, complement )
 import qualified Data.Foldable as F
 import           Data.Maybe (fromMaybe)
 import           Data.Semigroup
 import           Data.Typeable (Typeable)
-import           Data.Word ( Word32 )
+import           Data.Word ( Word8, Word16, Word32, Word64 )
 
 import qualified Codec.Binary.UTF8.String as UTF8 (decode)
 import qualified Control.Exception as X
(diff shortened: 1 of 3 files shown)


src/Data/LLVM/BitCode/Parse.hs
```diff
@@ -598,6 +600,28 @@ getTypeId n = do
     Just ix -> return ix
     Nothing -> fail ("unknown type alias " ++ show (ppLLVM llvmVlatest (llvmPP n)))
 
+-- | Convert a raw two's-complement value of the size specified in the first
+-- argument into a signed value.
+asSignedInt :: (Eq bitsize, Num bitsize, Enum bitsize)
+            => bitsize ->  Integer -> Integer
```
RyanGlScott, line 606: Is there a need for this function to be so polymorphic? As far as I can tell, it is only ever invoked in one place, where `bitsize` is instantiated to `Word32`. I would suggest just monomorphizing this to `asSignedInt :: Word32 -> Integer -> Integer`. (That should also allow GHC to specialize this code a bit more during optimization.)

src/Data/LLVM/BitCode/Parse.hs
```diff
@@ -598,6 +600,28 @@ getTypeId n = do
...
+asSignedInt nbits val =
+  let checkNeg x = if testBit x (fromEnum $ nbits - 1)
+                   then toInteger (complement x + 1) * (-1)
+                   else toInteger x
+  in case nbits of
+       8  -> checkNeg (fromInteger val :: Word8)
+       16 -> checkNeg (fromInteger val :: Word16)
+       32 -> checkNeg (fromInteger val :: Word32)
+       64 -> checkNeg (fromInteger val :: Word64)
+       _  -> val
```
RyanGlScott, line 616: Previously, we were only ever invoking `asSignedInt` in places where the bit size was known in advance (e.g., known to be 64), so I was reluctantly OK with this function only operating on these four bit sizes. Now that we are invoking `asSignedInt` to convert _all_ integer constants, however, this code is a lot more suspect. I think we really do need to make this work for all bit sizes, since it's entirely possible to write a constant like `i7 -1`, and we want to ensure that that is represented in the `llvm-pretty` AST as `ValInteger (-1)`, not `ValInteger 127`.

Thankfully, I don't think it would be that difficult to generalize this to work over any bit width. Taking inspiration from how the `bv-sized` library implements its [`asSigned`](https://hackage-content.haskell.org/package/bv-sized-1.0.7/docs/src/Data.BitVector.Sized.Internal.html#asSigned) function, I think the implementation of `asSignedInt` could simply be:

```hs
-- Precondition: @nbits > 0@
asSignedInt :: Word32 ->  Integer -> Integer
asSignedInt nbits val =
  let nbitsInt = fromIntegral @Word32 @Int nbits in
  if testBit val (nbitsInt - 1)
    then val - bit nbitsInt
    else val
```

RyanGlScott (commented):
Please make sure to add a test case involving negative integer constants to ensure that this fix works as expected. Per my comments elsewhere, it would also be good to ensure that different integer sizes are tested (e.g., include negative constants of type `i8` and `i7`).

PR #359, opened 2026-04-25
Add varargs disasm-test to ensure varargs compatibility.

kquick:


disasm-test/tests/varargs.ll
```diff
@@ -0,0 +1,191 @@
+; ModuleID = 'disasm-test/tests/varargs.c'
+source_filename = "disasm-test/tests/varargs.c"
+target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-i128:128-f80:128-n8:16:32:64-S128"
+target triple = "x86_64-unknown-linux-gnu"
+
+%struct.__va_list_tag = type { i32, i32, ptr, ptr }
```
RyanGlScott, line 6: Since this test case uses the `ptr` type, it won't work with pre-15 versions of LLVM. Please add a `varargs.pre-llvm15.ll` file to ensure that this test is skipped on those LLVM versions.

RyanGlScott (approved):

PR #360, opened 2026-04-27
Update llvm-disasm to use fixupOpaquePtrs to produce valid output.

kquick:


RyanGlScott:
What is the relationship between this PR and https://github.com/GaloisInc/llvm-pretty/pull/198? I ask since the comments you've added suggest that this is done to avoid pretty-printing things like `ptr*`, but shouldn't https://github.com/GaloisInc/llvm-pretty/pull/198 address this?

I'm not necessarily opposed to using `fixupOpaquePtrs` here, but doing so will have performance implications---I'd like to make sure we've consider all of the options first.

PR #364, opened 2026-04-30
Drop `tree-diff` dependency, use `llvm-diff` when available

### LINKED ISSUE #363: Reconsider dependency on `tree-diff`?
`disasm-test` depends on `tree-diff`:

https://github.com/GaloisInc/llvm-pretty-bc-parser/blob/3dd1be33bf481faf25017139ece848fb9029351a/llvm-pretty-bc-parser.cabal#L156

That's a surprisingly beefy package. It has 24 direct dependencies, and transitively depends on profunctors, adjunctions, kan-extensions, semigroupoids, distributive, comonad, and contravariant!

It's there for a good reason: making output of failing tests more readable. But I wonder if we could address this concern another way, perhaps by using [llvm-diff](https://github.com/llvm/llvm-project/blob/497d85067557c3ada0538928a99f5bcb38859fc7/llvm/docs/CommandGuide/llvm-diff.rst#L1), an LLVM-provided utility for diffing modules.

langston-barrett:
tree-diff is a surprisingly beefy package. It has 24 direct dependencies, and transitively depends on profunctors, adjunctions, kan-extensions, semigroupoids, distributive, comonad, and contravariant!

This commit replaces the use of `tree-diff` with a call to the external, llvm-provided `llvm-diff` utility, or `diff` if that's not available.

Fixes #363.

The following is a comparison of the output of these tools.

Modules:

```llvm
target triple = "unknown-unknown-unknown-unknown-"
declare void @foo()
declare void @bar()
```

```llvm
target triple = "unknown-unknown-unknown-unknown-"
declare void @foo()
declare void @baz()
```

Before — `tree-diff` (`prettyEditExprCompact`)

```
Differences (marked with + and - line prefixes:
Module {
  modSourceName = Nothing,
  modDeclares =
  [ Declare {
      decLinkage = Nothing,
      decVisibility = Nothing,
      decRetType = PrimType Void,
      decName = Symbol "foo",
      decArgs = [],
      decVarArgs = False,
      decAttrs = [],
      decComdat = Nothing},
    Declare {
      decLinkage = Nothing,
      decVisibility = Nothing,
      decName = Symbol -"bar" +"baz",
      decVarArgs = False,
      decComdat = Nothing,
      decRetType = ...}],
  modTriple = ...}
```

After — `llvm-diff`

```
llvm-diff output:
function @bar exists only in left module
function @baz exists only in right module
```

After — `diff -u` fallback (no `llvm-diff` on PATH)

```
diff -u output:
--- /tmp/demo1.ll  2026-04-30 12:50:53
+++ /tmp/demo2.ll  2026-04-30 12:50:53
@@ -1,3 +1,3 @@
 target triple = "unknown-unknown-unknown-unknown-"
 declare void @foo()
-declare void @bar()
+declare void @baz()
```

Arguably, `llvm-diff` produces nicer output than `tree-diff` anyway.


disasm-test/Main.hs
```diff
@@ -543,16 +544,31 @@ runAssemblyTest llvmVersion knownBugs sweet expct
...
+    tmp <- getTemporaryDirectory
+    let withLL name m f =
+          EX.bracket (openTempFile tmp name) (\(path, h) -> hClose h >> removeFile path) $ \(path, h) -> do
+            hPutStrLn h (show (ppLLVM llvmVlatest (llvmPP m)))
+            hClose h
+            f path
+    withLL "cmpASTs1.ll" ast1 $ \f1 ->
+      withLL "cmpASTs2.ll" ast2 $ \f2 -> do
+        irDiff <- do
+          mb <- findExecutable "llvm-diff"
```
kquick: Can you add a config override for finding this executable (see lines 166-195)?

disasm-test/Main.hs
```diff
@@ -543,16 +544,31 @@ runAssemblyTest llvmVersion knownBugs sweet expct
...
-cmpASTs ast1 ast2 = do
-  let d = ediff ast1 ast2
-      msg = "Differences (marked with + and - line prefixes:\n"
-            <> show (prettyEditExprCompact d)
-  liftIO $ assertBool msg $ ast1 == ast2
+cmpASTs ast1 ast2 =
+  when (ast1 /= ast2) $ liftIO $ do
+    tmp <- getTemporaryDirectory
+    let withLL name m f =
+          EX.bracket (openTempFile tmp name) (\(path, h) -> hClose h >> removeFile path) $ \(path, h) -> do
```
kquick: Should the cleanup operation include `hClose` or just do a `removeFile`?  The concern is cases where `hClose` causes an exception to be thrown.

disasm-test/Main.hs
```diff
@@ -543,16 +544,31 @@ runAssemblyTest llvmVersion knownBugs sweet expct
...
+            hPutStrLn h (show (ppLLVM llvmVlatest (llvmPP m)))
```
kquick: See the lines around 825-842: this should probably emit the AST with the version that corresponds to the llvm-tooling we are using to ensure it is processed correctly with `llvm-diff`.

disasm-test/Main.hs
```diff
@@ -543,16 +544,31 @@ runAssemblyTest llvmVersion knownBugs sweet expct
...
+            hClose h
+            f path
+    withLL "cmpASTs1.ll" ast1 $ \f1 ->
+      withLL "cmpASTs2.ll" ast2 $ \f2 -> do
+        irDiff <- do
```
kquick: I think `llvm-diff` has some built-in handling for metadata re-indexing immunity, so with this change, do we still need the `normalizeModule` functionality (line 789)?

disasm-test/Main.hs
kquick: And you might want to check packaging for typical distro's (RedHat, Ubuntu, and Alpine) to ensure that `llvm-tool` gets installed via the same package providing `llvm-as` and `llvm-dis`.  It probably is, in which case we probably don't also need to explicitly obtain the `llvm-diff` version for comparison and can just assume it is the same as `llvm-as`, but if you wanted to be thorough you could query its version and add it to the verification (~line 350).

kquick (changes requested):
I'm in favor of this generally because the `tree-diff` output is also not very distinct and I've found myself doing lots of scrolling back and forth through the output to try to see where the differences are.  Also, `llvm-diff` has some natural immunity to things like metadata re-indexing that would allow some simplification here.

disasm-test/Main.hs
langston-barrett: Hmm yeah I'm not sure. Perhaps it should do it in the other order (`removeFile path >> hClose h`)...?

disasm-test/Main.hs
langston-barrett: Done - added a `LLVMDiff` newtype with `IsOption` instance (matching the pattern of `LLVMAs`/`LLVMDis`/`Clang`) and a `--with-llvm-diff` option. It's stored in `TestState` and looked up via `gets llvmDiff` in `cmpASTs`.

disasm-test/Main.hs
langston-barrett: Fixed: the cleanup lambda now only calls `removeFile`

disasm-test/Main.hs
langston-barrett: Alpine does: https://pkgs.alpinelinux.org/package/edge/main/x86_64/llvm

RedHat does: https://rhel.pkgs.org/9/red-hat-ubi-appstream-x86_64/llvm-20.1.8-3.el9.x86_64.rpm.html

langston-barrett: Version verification added

disasm-test/Main.hs
langston-barrett: I don't really know much about this - how could I test it?

langston-barrett: Actually, I don't think this is relevant: `llvm-diff` is not used to *compare* the modules, just to output a pretty diff when they differ. So yes, I think normalization is still necessary.

langston-barrett:
Hmm, looks like we're missing `llvm-diff` in CI:
```
Tests
  Metadata
    test dedup metadata: OK
    test dedup metadata
      1:                 OK
      2:                 OK
      3:                 OK

All 4 tests passed (0.00s)
disasm-test: Unexpected version mismatch between clang, llvm-as, llvm-dis, llvm-diff
* llvm-as   version: llvm-as [12](https://github.com/GaloisInc/llvm-pretty-bc-parser/actions/runs/25189262540/job/73854740096?pr=364#step:8:13).0.1
* llvm-dis  version: llvm-dis 12.0.1
* llvm-diff version: llvm-diff [missing]
* clang     version: clang 12.0.1
```

RyanGlScott:
Ah, that's because we install LLVM utilities on a case-by-case basis here:

https://github.com/GaloisInc/llvm-pretty-bc-parser/blob/3dd1be33bf481faf25017139ece848fb9029351a/.github/workflows/build.yml#L39-L41

It should be straightforward to add `llvm-diff` to this list.

langston-barrett:
Ok this is ready again!

disasm-test/Main.hs
```diff
@@ -543,16 +563,51 @@ runAssemblyTest llvmVersion knownBugs sweet expct
...
+               Just 8 -> ppLLVM38 $ llvmPP m
+               o -> if maybe True (< 5) o
+                    then ppLLVM35 $ llvmPP m
+                    else ppLLVM38 $ llvmPP m
+        _ -> ppLLVM (fromEnum v) $ llvmPP m
+
+-- | Compare two ASTs to see if they are the same.  Fundamentally this is just
+-- done via (==) on the normalized ASTs.  On failure, serializes both modules to
+-- .ll files and runs llvm-diff for a structured diff; falls back to plain
+-- diff -u if llvm-diff is not available.
```
RyanGlScott, line 587: The Haddocks here talk about checking AST equality, which feels like a strange way to describe what this function is now doing (i.e., running `llvm-diff` on the pretty-printed modules).

disasm-test/Main.hs
```diff
@@ -543,16 +563,51 @@ runAssemblyTest llvmVersion knownBugs sweet expct
...
+    let ppForVersion = ppModuleForVersion llvmVersion
+    liftIO $ do
+      tmp <- getTemporaryDirectory
+      let withLL name m f =
+            EX.bracket (openTempFile tmp name) (\(path, _) -> removeFile path) $ \(path, h) -> do
+              hPutStrLn h (show (ppForVersion m))
+              hClose h
+              f path
+      withLL "cmpASTs1.ll" ast1 $ \f1 ->
+        withLL "cmpASTs2.ll" ast2 $ \f2 -> do
```
RyanGlScott: Similarly, the suffix "`ASTs`" here feels a bit odd.

disasm-test/Main.hs
```diff
@@ -340,22 +356,26 @@ main =  do
...
               , vcVersioning llvmAsVC == vcVersioning llvmDisVC
               , vcVersioning llvmAsVC == vcVersioning clangVC
+              , vcVersioning llvmAsVC == vcVersioning llvmDiffVC
               ]) $
     error $ unlines
-      [ "Unexpected version mismatch between clang, llvm-as and llvm-dis"
-      , "* llvm-as  version: " ++ showVC llvmAsVC
-      , "* llvm-dis version: " ++ showVC llvmDisVC
-      , "* clang    version: " ++ showVC clangVC
+      [ "Unexpected version mismatch between clang, llvm-as, llvm-dis, llvm-diff"
```
RyanGlScott: Perhaps this should use the word "or"? My understanding is that this will fail if there is a mismatch in at least one (but not necessarily all) of these tools:


```suggestion
      [ "Unexpected version mismatch between clang, llvm-as, llvm-dis, or llvm-diff"
```

disasm-test/Main.hs
```diff
@@ -543,16 +563,51 @@ runAssemblyTest llvmVersion knownBugs sweet expct
...
+    llvmVersion <- gets llvmVer
+    LLVMDiff diffExe <- gets llvmDiff
+    let ppForVersion = ppModuleForVersion llvmVersion
+    liftIO $ do
+      tmp <- getTemporaryDirectory
+      let withLL name m f =
+            EX.bracket (openTempFile tmp name) (\(path, _) -> removeFile path) $ \(path, h) -> do
+              hPutStrLn h (show (ppForVersion m))
+              hClose h
+              f path
```
RyanGlScott: I think [`withTempFile`](https://hackage.haskell.org/package/temporary-1.3/docs/System-IO-Temp.html#v:withTempFile) should allow you to easily create a temporary file without needing to manually `bracket` it.

disasm-test/Main.hs
```diff
@@ -543,16 +563,51 @@ runAssemblyTest llvmVersion knownBugs sweet expct
...
+      withLL "cmpASTs1.ll" ast1 $ \f1 ->
+        withLL "cmpASTs2.ll" ast2 $ \f2 -> do
+          mb <- findExecutable diffExe
+          irDiff <- case mb of
+            Just exe -> do
+              (_, out, err) <- readProcessWithExitCode exe [f1, f2] ""
+              return $ "llvm-diff output:\n" <> out <> err
+            Nothing -> do
+              (_, out, _) <- readProcessWithExitCode "diff" ["-u", f1, f2] ""
```
RyanGlScott, line 611: This file defines a [`callProc`](https://github.com/GaloisInc/llvm-pretty-bc-parser/blob/dac53f81f622b88a6fdd07e5ce47daf682b96b50/disasm-test/Main.hs#L967-L971) function that invokes a subprocess along with some extra logging if the `--details` flag is enabled. Probably worth using that here instead of `readProcessWithExitCode`.

RyanGlScott (approved):

disasm-test/Main.hs
langston-barrett: This function does check AST equality: https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/364/changes#diff-e36c67fc3c3b44684c2e7df8f15e605c5391aa5eee74cacba41090b0658db1bdR589

It also runs `llvm-diff` on the pretty-printed modules when the equality check fails. I think the Haddock explains this reasonably well, but I'm very open to suggestions on how to make it more clear!

disasm-test/Main.hs
langston-barrett: I don't think I can use `callProc`, because it doesn't return output and throws an exception on failure, which I don't think is appropriate here. I've added logging.

disasm-test/Main.hs
langston-barrett: Replaced with `mod1`/`mod2`.

disasm-test/Main.hs
langston-barrett: Done!

disasm-test/Main.hs
RyanGlScott: Oh, that's a bit odd. Since we're using `llvm-diff` anyway, is there a reason that we shouldn't just use that as the main source of truth for whether two modules are equal?

disasm-test/Main.hs
RyanGlScott: I didn't notice this discussion thread until now, but the idea is that `llvm-diff` can check for equality of two LLVM files modulo unnamed metadata indexing. For instance, if you have two files whose textual diff is the following:

```diff
$ diff -ru test1.ll test2.ll
--- test1.ll	2026-05-05 10:47:09
+++ test2.ll	2026-05-05 10:47:22
@@ -7,7 +7,7 @@
 define i32 @main() #0 !dbg !10 {
   %1 = alloca i32, align 4
   store i32 0, ptr %1, align 4
-  ret i32 0, !dbg !14
+  ret i32 0, !dbg !15
 }

 attributes #0 = { noinline nounwind optnone ssp uwtable(sync) "frame-pointer"="non-leaf" "no-trapping-math"="true" "probe-stack"="__chkstk_darwin" "stack-protector-buffer-size"="8" "target-cpu"="apple-m1" "target-features"="+aes,+altnzcv,+bti,+ccdp,+ccidx,+complxnum,+crc,+dit,+dotprod,+flagm,+fp-armv8,+fp16fml,+fptoint,+fullfp16,+jsconv,+lse,+neon,+pauth,+perfmon,+predres,+ras,+rcpc,+rdm,+sb,+sha2,+sha3,+specrestrict,+ssbs,+v8.1a,+v8.2a,+v8.3a,+v8.4a,+v8.5a,+v8a,+zcm,+zcz" }
@@ -30,4 +30,4 @@
 !11 = !DISubroutineType(types: !12)
 !12 = !{!13}
 !13 = !DIBasicType(name: "int", size: 32, encoding: DW_ATE_signed)
-!14 = !DILocation(line: 2, column: 3, scope: !10)
+!15 = !DILocation(line: 2, column: 3, scope: !10)
```

Then `llvm-diff` will report them as equal. Currently, we have to reindex the `Module`s to get the unnamed metadata index values to line up, but if we used `llvm-diff`, then we wouldn't need to.

disasm-test/Main.hs
langston-barrett: My biggest reason is to just keep compatbility with [how the tests worked before](https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/364/changes/cdab8b955b1b4c18331ef3e8e4fee0a0e358ae9b#diff-e36c67fc3c3b44684c2e7df8f15e605c5391aa5eee74cacba41090b0658db1bdL555). I'm not super confident that an empty diff from `llvm-diff` is the property that these tests want to be testing.

There's also a performance angle: `llvm-diff` is a subprocess, which is expensive. We generally expect these tests to pass. It's nice if they pass quickly, with a more expensive fallback to explain the failure only in cases where they do in fact fail.

RyanGlScott: Hm, OK. I think I'm on board with this for now, although we do run the risk of ending up in a situation where two `Module`s are unequal according to `(==)` but equal according to `llvm-diff` (e.g., two functions are equal modulo the naming of local variables). I suppose we can revisit this if it becomes a pain point in the future.

kquick (approved):

PR #365, opened 2026-05-26
Parse Windows SEH bitcode opcodes + operand bundles + real clang-cl artifact

AmeliaRose802:
### Summary

Adds bitcode parsing for the LLVM opcodes / blocks that `clang-cl /EHsc` (MSVC C++ EH) emits but that this parser previously rejected: SEH funclet opcodes, the `token` type, the personality function record, and operand bundles. Includes a real `clang-cl /EHsc` bitcode artifact under `disasm-test/bc_src_tests/` that exercises everything end-to-end.

Companion AST + pretty-printing changes live in [GaloisInc/llvm-pretty#201](https://github.com/GaloisInc/llvm-pretty/pull/201); the `llvm-pretty` submodule is bumped to match. The downstream `crucible-llvm` adapter that consumes the parsed AST lives in [GaloisInc/crucible#1839](https://github.com/GaloisInc/crucible/pull/1839).

### What's parsed

**SEH funclet opcodes (`FUNC_CODE_INST_*`):**

* `CATCHRET` (51) → `CatchRet`
* `CATCHPAD` (52) → `CatchPad`
* `CLEANUPPAD` (53) → `CleanupPad` (also reachable as a clause inside `LANDINGPAD`)
* `CLEANUPRET` (54) → `CleanupRet`
* `CATCHSWITCH` (56) → `CatchSwitch`

**`TYPE_CODE_TOKEN` (22):** parses the `token` type that catchswitch / catchpad / cleanuppad operate on.

**Personality function:** the `MODULE_CODE_FUNCTION` record's personality-function field is now read and surfaced on `Define`.

**`OPERAND_BUNDLE_TAGS_BLOCK_ID` (block 21):** each `OPERAND_BUNDLE_TAG` record (code 1) carries a tag string (e.g. `"funclet"`, `"deopt"`). We accumulate them into a module-level `Seq String` indexed in encounter order.

**`FUNC_CODE_OPERAND_BUNDLE` (opcode 55):** each record carries a tag index plus a sequence of `(value, type)` pairs encoded with `getValueTypePair`. Bundle records appear in the bitcode stream immediately before the call/invoke/callbr they attach to, so they're pushed onto a per-function `psPendingBundles` accumulator and consumed by the next `INVOKE` (13) / `CALL` (34) / `CALLBR` (57). The accumulator is reset on both entry and exit of `enterFunctionDef` so bundles cannot leak between functions.

[...]

PR #367, opened 2026-05-26
Replace notImplemented with comment stubs for SEH and token support

AmeliaRose802:
Replaces `notImplemented` errors with explanatory comment stubs for several SEH-related and token-type code paths in the bitcode parser. Allows bitcode that mentions these constructs (but does not necessarily exercise them on the verification target) to parse successfully so the downstream verifier can decide whether the constructs are reachable.

1 commit, low-risk error-message refactoring. Filed as draft for upstream review — the underlying SEH/token support PRs (`saw/seh-bitcode-only`, the AST-side `saw-token-type` on llvm-pretty) take a different, more complete approach. Open this PR mainly to surface the alternative for discussion.

### DIFF
--- a/src/Data/LLVM/BitCode/IR/Function.hs
+++ b/src/Data/LLVM/BitCode/IR/Function.hs
@@ -761,7 +761,7 @@ parseFunctionBlockEntry _ t d (fromEntry -> Just r) =
   -- [ptrty,ptr,cmp,new, align, vol,
   --  ordering, synchscope]
   37 -> label "FUNC_CODE_INST_CMPXCHG_OLD" $ do
-    notImplemented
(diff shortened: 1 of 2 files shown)


langston-barrett (changes requested):
It's not clear to me that these constructs are not verification-relevant. I would prefer actually parsing them to simply allowing them. In addition to a real implementation, a test-case would be welcome.

AmeliaRose802:
Closing in favor of #365, which takes the more complete approach (full bitcode decoding for the SEH opcodes + dependency on the AST/`Token` work in GaloisInc/llvm-pretty#201) rather than the comment-stub approach here.

PR #378, opened 2026-09-21
LLVM 23 support

RyanGlScott:
This is a collection of commits that allows `llvm-pretty-bc-parser` to parse most bitcode files using LLVM 23. Some changes are needed to support more new metadata fields (see also https://github.com/GaloisInc/llvm-pretty/pull/206, which adds support for these metadata fields on the `llvm-pretty` side), while other changes are needed to parse new module codes that `llvm-pretty` does not make meaningful use of. I also needed to update one test case (`dilocalvariable.ll`) to meet LLVM 23's stricter requirements about `DISubprogram` needing a non-null `type` field.

Fixes https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/371. Fixes https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/372. Fixes https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/374. Fixes https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/375.

PR #379, opened 2026-09-21
Update nix llvm references for post-22 support

kquick:


kquick:
@RyanGlScott I locally verified that this will work with LLVM 23 as well.  I can land this and you can rebase #378 and that should fix this nix CI.

RyanGlScott (approved):
Thank you!

